PR #9, opened 2017-02-17
Add tests for existing features People

gregdaynes:
These tests are for moving from Rails built in test runner to RSpec. 
Mostly a 1:1 of Rails tests, save for not checking `status: 200` page loads. They're tested when the template is rendered or redirected.


spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+    it "renders the template" do
+      expect(@response).to render_template(:index)
+    end
+    it "loads all of the people into @people" do
+      expect(@response.body.length).to be_truthy
+    end
+  end
+
+  describe "GET #show" do
+    before(:each) { @person = create(:person) }
```
graygilmore: Instead of before blocks you can utilize RSpec's `let`s.

```
let(:person) { FactoryGirl.create :person }
```

Generally we also prefer putting `FactoryGirl` in front of `.create` (or `.build`) so that we're being explicit in what is doing the creating (just like you've done `RSpec.describe` at the top of this file instead of `describe`).

graygilmore (changes requested):

spec/controllers/staff_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+      it "renders the template" do
+        put :update, params: { id: @staff, staff: FactoryGirl.attributes_for(:staff) }
+        expect(response).to redirect_to(@staff)
+      end
+    end
+
+    context "invalid attributes" do
+      it "should update staff" do
+        put :update, params: { id: @staff, staff: FactoryGirl.attributes_for(:staff, :no_pod) }
```
graygilmore: This isn't really an appropriate use of a trait since it's just a single change. This could be retwritten as:

```ruby
put :update, params: { id: @staff, staff: { pod: nil } }
```

As you can see we can also eliminate the use of `FactoryGirl.attributes_for`.

spec/controllers/staff_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+      it "re-runs staff#new" do
+        post :create, params: { staff: FactoryGirl.attributes_for(:staff, :invalid) }
+        expect(response).to render_template(:new)
+      end
+    end
+  end
+
+  describe "PATCH #update" do
+    before(:each) { @staff = create(:staff, pod: 2) }
```
graygilmore: I'd love for these tests to utilize `subject`. We could rewrite some of these like this:

```ruby
describe "PATCH #update" do
  subject { put :update, params: { id: @staff, staff: params }

  let(:params) {  { FactoryGirl.attributes_for(:person) } }

  it "blah blah" do
    subject
    expect(assigns(:staff)).to eq(@staff)
  end

  context "with invalid attributes" do
    let(:params) { { pod: nil } }

    # more tests
  end
end
```

spec/factories.rb
```diff
@@ -20,6 +20,14 @@
     pod 1
     started_at_company Date.new(2017,01,01)
     person
+
+    trait :invalid do
```
graygilmore: As mentioned above, traits are a great way to store complex changes to a factory and might not be the best use case here.

graygilmore (changes requested):
Great start for these tests. What I'd like to see is a refactor utilizing both `subject` and `let`.

spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  describe "GET #edit" do
+    subject { get :edit, params: { id: person.id } }
+    let(:person) { FactoryGirl.create :person }
+
+    it "renders the template" do
+      expect(subject).to render_template :edit
+    end
+    it "loads the person into @person" do
+      expect(assigns(person)).to eq(@person)
```
graygilmore: Does this pass? If it does this test might be broken!

spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    it "should located requested person" do
+      expect(assigns(person)).to eq(@person)
+    end
+
+    context "changing name" do 
+      let(:params) { { first_name: 'Jane' } }
+      it "should update person" do
+        subject
+        person.reload
+        expect(person.first_name).to eq('Jane')
```
graygilmore: One great way we could write this test to be super explicit about what we're testing would be:

```ruby
it "updates the person's name" do
  expect { subject }.to change { person.first_name }.from("Jessica").to("Jane")
end
```

spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      it "renders the template" do
+        expect(subject).to redirect_to(@person)
+      end
+    end
+
+    context "invalid attributes" do
+      let(:params) { { first_name: nil } }
+      it "shouldn't update person" do
+        subject
+        expect(person.first_name).to eq('John')
```
graygilmore: Similar to above we could write:

```ruby
it "does not update the person's name" do
  expect { subject }.to_not change { person.first_name }.from("John")
end
```

graygilmore (dismissed):

spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+RSpec.describe PeopleController, type: :controller do
+  describe "GET #index" do
+    subject { get :index }
+
+    it "renders the template" do
+      expect(subject).to render_template(:index)
+    end
+    it "loads all of the people into @people" do
+      expect(subject.body.length).to be_truthy
```
qr8r: This is some what a weird way to phrase this. 
I would test that the response code is a success.

qr8r (changes requested):

spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,111 @@
+# spec/controllers/people_controller_spec.rb
+require 'rails_helper'
+
+RSpec.describe PeopleController, type: :controller do
+  describe "GET #index" do
+    subject { get :index }
+
+    it "responds with 200" do
+      expect(subject.status).to eq(200)
```
qr8r: No need to change this, but I want to mention that The Response object here responds to a number of semantic 'status' methods like `:successful` allowing you to write something like `expect(subject).to be_successful` 

http://www.rubydoc.info/gems/rack/Rack/Response (included in ActionDispatch::Response)

spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      expect(subject).to render_template :edit
+    end
+  end
+
+  describe "POST #create" do
+    subject { post :create, params: { person: params } }
+    let(:params) { FactoryGirl.attributes_for(:person) }
+
+    it "changes person count by +1" do
+      expect { subject }.to change(Person, :count).by(1)
```
qr8r: Thats cool I hadnt seen the `change(Person, :count)` styntax before

spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    it "loads the person into @person" do
```
qr8r: Not sure this is the best description for the spec. Seems like an artifact from when you were testing `assigns`

spec/controllers/people_controller_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      it "should update person" do
+        expect { subject }
+          .to change { person.reload.first_name }
+          .from('John')
+          .to('Jane')
+      end
+
+      it "renders the template" do
+        expect(subject).to redirect_to(@person)
```
qr8r: Im not sure I like referencing this instance variable directly. Im assuming the `let` statement internally creates this instance variable... would probably be better just to call the `:person` let statement

qr8r (changes requested):

qr8r (approved):

gregdaynes:
Going to update the Model Tests to use FactoryGirl. Thank you @natyork for pointing it out.

PR #11, opened 2017-02-20
Remove old Models, Views, Controllers and Specs

gregdaynes:
Changing direction of the application to not have staff members, this PR is to remove related object from the codebase.


qr8r (approved):
Removes models we dont need :+1:

PR #12, opened 2017-02-20
Feature/add rubocop

qr8r:
Install rubocop for pedantic ruby review

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,6 +30,9 @@ Metrics/BlockLength:
   ExcludedMethods:
     - 'describe'
     - 'context'
+    - 'define'
+    - 'trait'
+    - 'factory'
 
 Metrics/LineLength:
   IgnoredPatterns:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,36 @@
+AllCops:
+  Exclude:
+    - 'db/**/*'
+    - 'config/**/*'
+    - 'Gemfile*'
+    - 'bin/**/*'
+    - 'test/**/*'
+    - 'Rakefile'
+    - 'spec/spec_helper.rb'
+    - 'spec/rails_helper.rb'
+    - 'vendor/**/*'
+    - 'Guardfile'
+
+Rails:
+  Enabled: true
+
+Style/FrozenStringLiteralComment:
+  Enabled: false
+
+Style/Documentation:
+  Enabled: false
+
+Style/StringLiterals:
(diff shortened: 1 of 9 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,36 @@
...
+    - 'bin/**/*'
+    - 'test/**/*'
+    - 'Rakefile'
+    - 'spec/spec_helper.rb'
+    - 'spec/rails_helper.rb'
+    - 'vendor/**/*'
+    - 'Guardfile'
+
+Rails:
+    Enabled: true
```
graygilmore: Would this double indent pass Rubocop!? 😉

qr8r: Hahahah

qr8r:
Jenkins test this please

qr8r:
Jenkins test this please

qr8r:
Jenkins test this please

jacquesporveau:
Rubocop style errors are fixed however Brendan is going to be doing some further work with integrating Jenkins thus the CI is still failing.

qr8r:
Jenkins test this please

qr8r:
Jenkins test this please

app/controllers/people_controller.rb
```diff
@@ -47,6 +47,9 @@ def destroy
   private
 
   def person_params
-    params.require(:person).permit(:first_name, :last_name, :email, :password, :password_confirmation, :date_of_birth, :gender)
+    params.require(:person).permit(
+      :first_name, :last_name, :email, :password,
+      :password_confirmation, :date_of_birth, :gender
```
gregdaynes: Any reason for this formatting over
:first_name,
:last_name,
...

jacquesporveau: I prefer this style personally but I'll have a look at the style guide and make the appropriate changes.

qr8r: Haha I actually agree with this :P I didnt actually review the Rubocop corrections :(

qr8r: The advantage of listing one arg per line is that if a single param changes in the future the git diff will be much easier to read. 

if there are multiple things on one line, and one item in the list changes, git highlights the whole line making it harder to identify what exactly on that line changed.

Ultimately, this is the kind of thing that we should rely on rubocop for. So long as we agree on which style we prefer, we can set up rubocop to enforce it

gregdaynes (approved):

Issue #2, opened 2017-02-17
Event Creation


graygilmore:
As an event organizer,
I would like to be able to create events.

## Acceptance Criteria

- I should be able to create an event with the following attributes:
  - Date
  - Location
  - Name
- All fields are mandatory
- I should see error messages if any of the fields are blank

PR #16, opened 2017-02-21
Event creation 

gregdaynes:
added ability to create events with Name, Date, and Location

added views for creating, and viewing created event
added view partials for form and error reporting

Specs for Controller + Model

#2

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   resources :people
+  resources :events, only: [:show, :create, :new]
 
   root 'welcome#index'
 end
--- a/db/migrate/20170221003240_create_events.rb
+++ b/db/migrate/20170221003240_create_events.rb
@@ -1,0 +1,11 @@
+class CreateEvents < ActiveRecord::Migration[5.0]
+  def change
+    create_table :events do |t|
+      t.date :date, :null => false
+      t.string :location, :null => false
+      t.string :name, :null => false
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20170221003240_create_events.rb
+++ b/db/migrate/20170221003240_create_events.rb
@@ -1,9 +1,9 @@
 class CreateEvents < ActiveRecord::Migration[5.0]
   def change
     create_table :events do |t|
-      t.date :date, :null => false
-      t.string :location, :null => false
-      t.string :name, :null => false
+      t.date :date, null: false
+      t.string :location, null: false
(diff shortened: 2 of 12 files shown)


graygilmore (changes requested):
Show page is empty :wink:

config/routes.rb
```diff
@@ -1,5 +1,7 @@
 Rails.application.routes.draw do
   resources :people
+  resources :presentations
```
graygilmore: I think this line is in the wrong PR.

db/migrate/20170221003240_create_events.rb
```diff
@@ -0,0 +1,11 @@
+class CreateEvents < ActiveRecord::Migration[5.0]
+  def change
+    create_table :events do |t|
+      t.date :date
```
graygilmore: Because we never want these to be blank values we should also include a `null: false` for each of them. The Rails validations are great but if we really don't want these to be blank we should have DB validations as well.

spec/models/event_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+RSpec.describe Event, type: :model do
+  subject { described_class.new(event) }
+
+  describe 'valid model' do
+    let(:event) { FactoryGirl.attributes_for(:event) }
+    it { is_expected.to be_valid }
+  end
+
+  describe 'name' do
+    context 'when nil' do
```
graygilmore: If we have a context for one direction we'll want an example for the other direction.

graygilmore (changes requested):

graygilmore (approved):

graygilmore (approved):
@jacquesporveau if this looks good to you, merge away!

db/migrate/20170221003240_create_events.rb
```diff
@@ -0,0 +1,11 @@
+class CreateEvents < ActiveRecord::Migration[5.0]
+  def change
+    create_table :events do |t|
+      t.date :date, :null => false
+      t.string :location, :null => false
+      t.string :name, :null => false
```
jacquesporveau: It's weird that Rubocop isn't picking this up but this is the old hash syntax.

`:null => false` becomes `null: false`

gregdaynes: Right, right. 
RuboCop is currently ignoring the db folder with db/**/*

jacquesporveau:
I went ahead and made the changes that I requested. New hash syntax.

PR #18, opened 2017-02-22
Feature/event notification

jacquesporveau:
- Users should be able to configure how many days before a presentation the presenter is notified of their talk.
 - The configuration should apply to all upcoming presentations

### DIFF
--- a/app/jobs/event_notification_job.rb
+++ b/app/jobs/event_notification_job.rb
@@ -1,0 +1,10 @@
+class EventNotificationJob < ApplicationJob
+  def self.perform(event)
+    event.presentations.joins(:person).each do |presentation|
+      PersonMailer.presentation_notification(
+        presentation.person,
+        presentation.event.date
+      ).deliver
+    end
+  end
+end
--- a/app/mailers/person_mailer.rb
+++ b/app/mailers/person_mailer.rb
@@ -1,0 +1,14 @@
+class PersonMailer < ApplicationMailer
+  include ActionView::Helpers::DateHelper
+  def presentation_notification(presenter, event_date)
+    @presenter = presenter
+    @event_date = event_date.strftime('%a %b %d')
+    @notification_period = distance_of_time_in_words(
+      Event::NOTIFICATION_PERIOD
+    )
+    mail(
+      to: presenter.email,
+      subject: "You have a presentation in #{@notification_period}"
+    )
+  end
+end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,5 +1,6 @@
 # app/models/event.rb
 class Event < ApplicationRecord
+  NOTIFICATION_PERIOD = 1.day
   has_many :presentations
 
   validates :date,
@@ -10,4 +11,7 @@ class Event < ApplicationRecord
             presence: true
 
   scope :upcoming, -> { where('date >= ?', Time.zone.today) }
+  scope :requiring_notification, -> do
+    where('date = ?', Date.current - NOTIFICATION_PERIOD)
+  end
 end
--- a/app/views/person_mailer/presentation_notification.html.erb
+++ b/app/views/person_mailer/presentation_notification.html.erb
@@ -1,0 +1,12 @@
+<DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <h1>Greetings <%= @presenter.first_name if @presenter.first_name %></h1>
+    <p>This is a reminder for your upcoming presentation on
+      <%= @event_date %>
+    </p>
+  </body>
+</html>
--- a/lib/tasks/send_presentation_notifications.rake
+++ b/lib/tasks/send_presentation_notifications.rake
@@ -1,0 +1,8 @@
+require 'rake'
+
+desc 'Sends email notifications to all presenters of upcoming events'
+task send_presentation_notifications: :environment do
+  Event.requiring_notification.each do |event|
+    EventNotificationJob.perform_later(event)
+  end
+end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -3,11 +3,25 @@ FactoryGirl.define do
   factory :person do
     first_name 'John'
     last_name 'Doe'
-    email "john.doe#{Random.rand(1000)}@example.com"
+    email { "john.doe#{Random.rand(1000)}@example.com" }
     password 'abc123'
     password_confirmation 'abc123'
     date_of_birth Date.new(1970, 1, 1)
     gender 'Computer'
+  end
+
+  factory :event do
+    name 'Show & Tell'
+    location 'Board Room'
+    date Date.new(2017, 2, 24)
+
+    trait :requiring_notification do
+      date Time.zone.today - Event::NOTIFICATION_PERIOD
+    end
+
+    trait :not_requiring_notification do
+      date Time.zone.today + Event::NOTIFICATION_PERIOD
+    end
   end
 
   factory :presentation do
@@ -16,11 +30,9 @@ FactoryGirl.define do
     duration 900
     person
     event
-  end
 
-  factory :event do
-    name 'Show & Tell'
-    location 'Board Room'
-    date Date.new(2017, 2, 24)
+    trait :requiring_notification do
+      event { FactoryGirl.build(:event, :requiring_notification) }
+    end
   end
 end
--- a/spec/jobs/event_notification_job_spec.rb
+++ b/spec/jobs/event_notification_job_spec.rb
@@ -1,0 +1,25 @@
+require 'rails_helper'
+
+describe EventNotificationJob do
+  subject(:performed_job) { described_class.perform(event) }
+  let(:event) { FactoryGirl.create(:event, :requiring_notification) }
+  before { FactoryGirl.create(:presentation, event: event) }
+
+  describe '.perform' do
+    context 'when there are multiple presentations for an event' do
+      before { FactoryGirl.create(:presentation, event: event) }
+      it 'sends an email to the presenter of those presentations' do
+        expect { performed_job }
+          .to change { ActionMailer::Base.deliveries.count }
+          .by(event.presentations.length)
+      end
+    end
+    context 'when there are no presentations for an event' do
+      before { event.presentations = [] }
+      it 'sends zero emails' do
+        expect { performed_job }
+          .not_to change { ActionMailer::Base.deliveries.count }
+      end
+    end
+  end
+end
--- a/spec/models/event_spec.rb
+++ b/spec/models/event_spec.rb
@@ -57,4 +57,29 @@ RSpec.describe Event, type: :model do
       it { is_expected.to_not include(event_past) }
     end
   end
+
+  describe '.requiring_notification' do
+    subject(:events_requiring_notification) do
+      described_class.requiring_notification
+    end
+    context 'when there are events aligning with the NOTIFICATION_PERIOD' do
+      let(:required_notif) do
+        FactoryGirl.create(:event, :requiring_notification)
+      end
+      let(:unrequired_notif) do
+        FactoryGirl.create(:event, :not_requiring_notification)
+      end
+      it 'returns an array of Events requiring notification' do
+        expect(events_requiring_notification).to include(required_notif)
+      end
+      it 'doesnt include Events that arent requiring notification' do
+        expect(events_requiring_notification).not_to include(unrequired_notif)
+      end
+    end
+    context 'when there are no events aligning with the NOTIFICATION_PERIOD' do
+      it 'returns an empty array' do
+        expect(events_requiring_notification).to be_empty
+      end
+    end
+  end
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,3 +46,4 @@ Metrics/MethodLength:
 
 Metrics/AbcSize:
   Max: 25
+  Enabled: false
--- a/spec/mailers/person_mailer_spec.rb
+++ b/spec/mailers/person_mailer_spec.rb
@@ -1,0 +1,14 @@
+require 'rails_helper'
+
+describe PersonMailer do
+  describe '.presentation_notification' do
+    let(:person) { FactoryGirl.create(:person) }
+    subject do
+      described_class.presentation_notification(person, Date.current).deliver
+    end
+
+    it 'greets the presenter' do
+      expect(subject.body).to include("Greetings")
+    end
+  end
+end


app/models/event.rb
```diff
@@ -1,9 +1,14 @@
...
 class Event < ApplicationRecord
+  NOTIFICATION_PERIOD = 1.day
   validates :date,
             presence: true
   validates :location,
             presence: true
   validates :name,
             presence: true
+
+  def self.requiring_notification
```
qr8r: Consider making this an AR scope instead of a class method.

spec/models/event_spec.rb
```diff
@@ -41,4 +41,30 @@
       it { is_expected.to_not be_valid }
     end
   end
+
+  describe '.requiring_notification' do
+    context 'when there are events alrigning with the NOTIFICATION_PERIOD' do
```
qr8r: Typo?

spec/models/event_spec.rb
```diff
@@ -41,4 +41,30 @@
       it { is_expected.to_not be_valid }
     end
   end
+
+  describe '.requiring_notification' do
```
qr8r: Your spec should have a `subject` defined. the subject helper takes an optional symbol param allowing you to make more semantic expectations 
```ruby
subject(:smarties) { ... }

it 'sdfsdfsdf' do
  expect(:smarties).to be_better_than :m_and_ms
end
```

spec/models/event_spec.rb
```diff
@@ -41,4 +41,30 @@
...
+      it 'returns an array of Events requiring notification' do
+        expect(Event.requiring_notification).to include(required_notif)
+      end
+      it 'doesnt include Events that arent requiring notification' do
+        expect(Event.requiring_notification).not_to include(unrequired_notif)
+      end
+    end
+    context 'when there are no events aligning with the NOTIFICATION_PERIOD' do
+      it 'returns an empty array' do
+        expect(Event.requiring_notification.empty?).to eq(true)
```
qr8r: Rspec allows you to dynamically use predicate methods with the `be` helper. 

`expect([]).to be_empty`

app/models/event.rb
```diff
@@ -1,9 +1,14 @@
 # app/models/event.rb
```
qr8r, line 1: This commit needs a commit message explaining why this change is required.

app/helpers/email_helper.rb
```diff
@@ -0,0 +1,30 @@
+module EmailHelper
+  class SendPresentationNotifications
```
qr8r: This class strikes me as a Job. 

Consider inheriting from ApplicationJob

app/helpers/email_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+      # end
+      #
+      # When you uncomment this code make sure to update the specs in
+      # spec/helpers/email_helper_spec.rb
+      #
+      PersonMailer.presentation_notification(presenter).deliver
+    end
+  end
+
+  def self.send_presentation_notifications
```
qr8r: This method might not be needed if you inherit from ApplicationJob

app/jobs/event_notification_job.rb
```diff
@@ -0,0 +1,6 @@
+class EventNotificationJob < ApplicationJob
+  helper EmailHelper
```
qr8r: Unneeded helper statement. The EmailHelper has no public api, so declaring it as a helper here does nothing

app/jobs/event_notification_job.rb
```diff
@@ -0,0 +1,6 @@
...
+  def perform
```
qr8r: I think this class should be doing what your Email helper is doing.

qr8r (changes requested):

app/models/event.rb
jacquesporveau: Hey I didn't know about that syntax. Very cool. Thanks for the tip.

spec/models/event_spec.rb
```diff
@@ -41,4 +41,31 @@
       it { is_expected.to_not be_valid }
     end
   end
+
+  describe '.requiring_notification' do
+    subject(:events_requiring_notification) { Event.requiring_notification }
```
qr8r: No need to change this, is good to know that when you are referencing the class that you are testing in RSpec, you can use the `described_class` helper.

Makes it easier to change the name of the class if you ever had to

qr8r (commented):
Looks good. Will approve once #19 is merged

spec/models/event_spec.rb
jacquesporveau: Will keep that in mind.

spec/models/event_spec.rb
```diff
@@ -41,4 +41,31 @@
...
+      end
+      let(:unrequired_notif) do
+        FactoryGirl.create(
+          :event, date: Time.zone.today + Event::NOTIFICATION_PERIOD
+        )
+      end
+      it 'returns an array of Events requiring notification' do
+        expect(events_requiring_notification).to include(required_notif)
+      end
+      it 'doesnt include Events that arent requiring notification' do
```
gregdaynes, line 75: Double negative - reads funny.
it 'only includes events that require notification' or something? minor, and doesn't really matter.

app/models/event.rb
```diff
@@ -1,9 +1,14 @@
...
 class Event < ApplicationRecord
+  NOTIFICATION_PERIOD = 1.day
```
gregdaynes, line 3: Unfamiliar with Rails, but this would be beneficial to be defined in a constants/config file for easier future changes.

`config/application.rb`
then for dev/testing purposes, overload it in `confing/environemnts/`

gregdaynes: Nevermind. I guess that's against common Rails practive.

jacquesporveau: Totally. We are adding the Figaro gem in the oauth work that I am doing. Once that is configured I think that's a great idea.

app/jobs/event_notification_job.rb
```diff
@@ -0,0 +1,20 @@
+class EventNotificationJob < ApplicationJob
+  def self.perform
+    presenter = {
+      first_name: 'Dennis',
+      last_name: 'Marchand',
+      email: 'dennis@urbantherapy.ca'
+    }
+
+    # Event.requiring_notification.each do |event|
```
graygilmore: I think we should remove the commented out code. Instead, I'd prefer to add a `#TODO` comment to `presenter` above, detail that this is a stubbed model for now.

app/views/person_mailer/presentation_notification.html.erb
```diff
@@ -0,0 +1,9 @@
+<DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <p>Hello world</p>
```
graygilmore: I think we can add actual content here since we're just stubbing out the model that we're going to be getting anyways.

app/mailers/person_mailer.rb
```diff
@@ -0,0 +1,6 @@
+class PersonMailer < ApplicationMailer
+  def presentation_notification(presenter)
+    @presenter = presenter
+    mail(to: presenter[:email], subject: 'Hello world')
```
graygilmore: Maybe `subject: "You have a presentation in #{n} days"`?

graygilmore (changes requested):

app/jobs/event_notification_job.rb
jacquesporveau: With @gregdaynes stuff being merged I think this is ready to code now so instead of leaving this stubbed I'll just finish it.

jacquesporveau: This has been completed in 1b60260

app/jobs/event_notification_job.rb
```diff
@@ -0,0 +1,9 @@
+class EventNotificationJob < ApplicationJob
+  def self.perform
```
graygilmore: When does this run?

graygilmore (changes requested):
I would like to see the last commit squashed into the one before it.

app/jobs/event_notification_job.rb
jacquesporveau: Right now it doesn't we haven't added in Redis or another task runner yet.

qr8r: I will add stories to implement a scheduling provider and a background processing queue provider.
We wont actually be running this stuff until we set up a staging environment

For now this only requires a rake task to manually run the job from the console. In the Development environment by default Active job will run jobs inline, until we implement a background processing queue

jacquesporveau: rake task has been introduced in dfb704af07ab8701537e45a13d65bef87cdf8dd0

qr8r:
What does `RFR` mean in the description? 
Ready for review.

app/jobs/event_notification_job.rb
```diff
@@ -0,0 +1,13 @@
...
+    Event.requiring_notification.each do |event|
+      event_date = event.date
```
qr8r: Unnecessary variable assignment. `event`  is still available in the nested loop and typing `event_date` doesn't save much space.

app/jobs/event_notification_job.rb
```diff
@@ -0,0 +1,13 @@
+class EventNotificationJob < ApplicationJob
+  def self.perform
+    Event.requiring_notification.each do |event|
```
qr8r: Probably want to eager load presentations and person here so that they are all fetch in the same query. This way we avoid creating an N+1 query to the database. 

If there were a large number of events involved, we might want to consider using the `find_each` AR method rather than a simple each. but for now its probably fine

spec/factories.rb
```diff
@@ -3,7 +3,7 @@
   factory :person do
     first_name 'John'
     last_name 'Doe'
-    email "john.doe#{Random.rand(1000)}@example.com"
+    sequence(:email) { |n| "john.doe#{n}@example.com" }
```
qr8r: What is this being changed? There is a unique validation on emails making sequences a little clunky if you were ever going to use the factories in the dev environment

spec/jobs/event_notification_job_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+  context 'when there are no events requiring notification' do
+    let!(:unrequired_notif) do
+      FactoryGirl.create(
+        :event, date: Date.current + Event::NOTIFICATION_PERIOD
+      )
+    end
+    it 'sends zero emails' do
+      expect(events_requiring_notification.length).to eq(0)
+      expect { job.perform }
```
qr8r: described class.perform should probably be your subject since thats what you are "expecteing" in both tests

app/jobs/event_notification_job.rb
```diff
@@ -0,0 +1,13 @@
+class EventNotificationJob < ApplicationJob
```
qr8r, line 1: This commit is too large. The mailer and the job should be in separate commits

lib/tasks/send_presentation_notifications.rake
```diff
@@ -0,0 +1,5 @@
+require 'rake'
+
+task send_presentation_notifications: :environment do
+  EventNotificationJob.perform
```
qr8r: So it took me until now to realize this, but our job doesnt take any arguments. Which means we can't take advantage of splitting our event notifications out into multiple processes to make them faster.

I propose making EventNotificationJob#perform accept an event as an argument. 
It will be the rake tasks responsibility to find all events that need to be notified, and then initialize a new job for every event.

That way when we implement our background processing. The queue can run multiple jobs at the same time, instead of having to notify every event from within the same job

qr8r (changes requested):
Getting there!

spec/factories.rb
jacquesporveau: This doesn't work in the test environment. If you attempt to use this style of get a Validation failed: Email has already been taken error. I've googled a bunch and tried to debug it and every source I find recommends doing things this way so I tried it and it solved the test environment issues I was having.

app/jobs/event_notification_job.rb
jacquesporveau: Got it!

graygilmore (approved):

spec/factories.rb
jacquesporveau: Alright found the solution for this.

https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#dynamic-attributes

jacquesporveau:
Weird. Had to reset branch set to merge into to re-check for merge conflicts as it was picking up some remotely but not locally.

app/models/event.rb
```diff
@@ -10,4 +11,8 @@ class Event < ApplicationRecord
             presence: true
 
   scope :upcoming, -> { where('date >= ?', Time.zone.today) }
+  scope :requiring_notification, -> do
+    where('date = ?', Date.current - NOTIFICATION_PERIOD)
+      .includes(presentations: [:person])
```
qr8r: The `[]` around person here is not needed.

In this case I wouldn't preload the associations here (unless the scope itself needed it for some reason). you take this scope from being 1 query to 3, and you might not always want the associated records.

Better to leave the `includes` call to the person who will actually be using the scope

spec/factories.rb
```diff
@@ -3,24 +3,36 @@
   factory :person do
     first_name 'John'
     last_name 'Doe'
-    email "john.doe#{Random.rand(1000)}@example.com"
+    email { "john.doe#{Random.rand(1000)}@example.com" }
```
qr8r, line 6: :+1:

lib/tasks/send_presentation_notifications.rake
```diff
@@ -0,0 +1,7 @@
+require 'rake'
+
+task send_presentation_notifications: :environment do
+  Event.requiring_notification.each do
+    EventNotificationJob.perform(event)
```
qr8r: You probably want perform_later here. When there is no processing background in place, rails will always perform jobs inline, even if you asked it to `perform_later`.

This way we dont need to remember to change this when we add a processing background

qr8r (changes requested):
Almost!!!

app/models/event.rb
jacquesporveau: Hmmmm right I see where you are coming from however the instance in which I need to do that preloading is inside a rake task. I am not sure if we want that much logic in there?

```
require 'rake'

task send_presentation_notifications: :environment do
  Event.requiring_notification.each do
    EventNotificationJob.perform(event)
  end
end
```

That's what it's looking like right now. I could put the eager load in there but I just remember you mentioning to stay away from logic inside rake tasks.

spec/factories.rb
jacquesporveau: YUS

lib/tasks/send_presentation_notifications.rake
jacquesporveau: Is that some Rails magic or something? I have a method defined as perform so I assumed that's what I had to call. Does it know about perform_later? 

Just looked up the docs. Yes it does? Weird.

app/models/event.rb
jacquesporveau: Alright got this one figured out. The scope in the event model is just for itself and in the event job I have...
```
class EventNotificationJob < ApplicationJob
  def self.perform(event)
    event.presentations.joins(:person).each do |presentation|
      PersonMailer.presentation_notification(
        presentation.person,
        presentation.event.date
      ).deliver
    end
  end
end
```

jacquesporveau: Tried this in console and it grabs only presentations that have a person as well as loading the people in memory.

lib/tasks/send_presentation_notifications.rake
```diff
@@ -0,0 +1,7 @@
+require 'rake'
+
+task send_presentation_notifications: :environment do
```
qr8r: Have you been able to run this rake task from the console?

spec/factories.rb
```diff
@@ -3,24 +3,36 @@
...
     date_of_birth Date.new(1970, 1, 1)
     gender 'Computer'
   end
 
+  factory :event do
+    name 'Show & Tell'
+    location 'Board Room'
+    date Date.new(2017, 2, 24)
+
+    trait :requiring_notification do
```
qr8r, line 18: :+1: Came in handy when I wanted to test this locally. Good call

qr8r (changes requested):

lib/tasks/send_presentation_notifications.rake
jacquesporveau: Sry man that's my bad. I should have tested this more in depth. Made the missing var change.

qr8r (approved):

PR #19, opened 2017-02-23
Presentation creation

gregdaynes:
https://www.pivotaltracker.com/story/show/140277459

Adds Feature to create a presentation, referencing to an existing event and person (or presenter if not an existing person)

This also further removes legacy relations of has_many on presentation/people

### DIFF
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -6,4 +6,28 @@ class PresentationsController < ApplicationController
   def new
     @presentation = Presentation.new
   end
+
+  def create
+    @presentation = Presentation.new(presentation_params)
+
+    if @presentation.save
+      flash[:notice] = 'Presentation was successfully created'
+      redirect_to @presentation
+    else
+      render 'new'
+    end
+  end
+
+  private
+
+  def presentation_params
+    params.require(:presentation).permit(
+      :presenter,
+      :topic,
+      :duration,
+      :description,
+      :person_id,
+      :event_id
+    )
+  end
 end
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -1,4 +1,8 @@
 class PresentationsController < ApplicationController
+  def show
+    @presentation = Presentation.find(params[:id])
+  end
+
   def new
     @presentation = Presentation.new
   end
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -1,0 +1,5 @@
+class PresentationsController < ApplicationController
+  def new
+    @presentation = Presentation.new
+  end
+end
--- a/app/models/presentation.rb
+++ b/app/models/presentation.rb
@@ -1,4 +1,24 @@
 class Presentation < ApplicationRecord
-  has_many :person_presentations
-  has_many :people, through: :person_presentations
+  belongs_to :person, optional: true
+  belongs_to :event
+
+  validates :topic,     presence: true
+  validates :duration,  presence: true
+  validates :person,    presence: true, unless: :presenter
+  validates :presenter, presence: true, unless: :person
+  validates :event,     presence: true
+
+  validate  :person_or_presenter
+
+  def presenter_name
+    person ? person.first_name : presenter
+  end
+
+  private
+
+  def person_or_presenter
+    return unless person.present? && presenter.present?
+
+    errors.add(:presenter, 'Can only have a Person or a Presenter')
+  end
 end
--- a/app/views/presentations/_errors.html.erb
+++ b/app/views/presentations/_errors.html.erb
@@ -1,0 +1,13 @@
+<% if @presentation.errors.any? %>
+  <div>
+    <h2>
+      <%= pluralize(@presentation.errors.count, "error") %>
+      Prohibited this presentation from being created
+    </h2>
+    <ul>
+      <% @presentation.errors.full_messages.each do |msg| %>
+        <li><%= msg %></li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
--- a/app/views/presentations/_form.html.erb
+++ b/app/views/presentations/_form.html.erb
@@ -1,0 +1,35 @@
+<%= form_for @presentation do |f| %>
+  <%= render 'errors' %>
+
+  <p>
+    <%= f.label :topic %><br>
+    <%= f.text_field :topic %>
+  </p>
+
+  <p>
+    <%= f.label :description %><br>
+    <%= f.text_field :description %>
+  </p>
+
+  <p>
+    <%= f.label :duration %><br>
+    <%= f.text_field :duration %> Seconds
+  </p>
+
+  <p>
+    <%= f.label :event %><br>
+    <%= collection_select(:presentation, :event_id, Event.all, :id, :name, prompt: true) %>
+  </p>
+
+  <p>
+    <%= f.label :presenter %><br>
+    <%= collection_select(:presentation, :person_id, Event.all, :id, :name, include_blank: true) %>
+    or
+    <%= f.text_field :presenter, placeholder: 'Enter guest name' %>
+  </p>
+
+  <p>
+    <%= f.submit %>
+  </p>
+
+<% end %>
--- a/app/views/presentations/new.html.erb
+++ b/app/views/presentations/new.html.erb
@@ -1,0 +1,1 @@
+<h1>New Presentation</h1>
--- a/app/views/presentations/new.html.erb
+++ b/app/views/presentations/new.html.erb
@@ -1,1 +1,3 @@
 <h1>New Presentation</h1>
+
+<%= render 'form' %>
--- a/spec/controllers/presentations_controller_spec.rb
+++ b/spec/controllers/presentations_controller_spec.rb
@@ -9,10 +9,47 @@ RSpec.describe PresentationsController, type: :controller do
   end
 
   describe 'GET #show' do
-    subject { get :show, params: { id: 1 } }
+    subject { get :show, params: { id: presentation.id } }
     let!(:presentation) { FactoryGirl.create(:presentation) }
 
     it { is_expected.to render_template :show }
     it { is_expected.to be_successful }
+  end
+
+  describe 'POST #create' do
+    subject { post :create, params: { presentation: params } }
+    let!(:person) { FactoryGirl.create(:person) }
+    let!(:event) { FactoryGirl.create(:event) }
+
+    context 'valid params' do
+      let(:params) do
+        FactoryGirl.attributes_for(
+          :presentation,
+          person_id: person.id,
+          event_id: event.id
+        )
+      end
+
+      it { is_expected.to redirect_to Presentation.last }
+
+      it 'is expected to be successful' do
+        expect(response).to be_success
+      end
+
+      it 'is expected to change presentation count +1' do
+        expect { subject }.to change(Presentation, :count).by(1)
+      end
+    end
+
+    context 'invalid params' do
+      let(:params) { FactoryGirl.attributes_for(:presentation) }
+
+      it { is_expected.to render_template :new }
+      it { is_expected.to be_successful }
+
+      it 'is expected to not change presentation count' do
+        expect { subject }.to_not change(Presentation, :count)
+      end
+    end
   end
 end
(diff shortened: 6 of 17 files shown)


app/models/presentation.rb
```diff
@@ -1,4 +1,17 @@
...
-  has_many :people, through: :person_presentations
+  belongs_to :person, optional: true
+  belongs_to :event
+
+  validates :topic,     presence: true
+  validates :duration,  presence: true
+  validates :presenter, presence: true, unless: :person
+  validates :person,    presence: true, unless: :presenter
+
+  def presenter_name
```
graygilmore: We should put this method under test, too!

spec/models/presentation_spec.rb
```diff
@@ -1,5 +1,109 @@
+# spec/models/presentation_spec.rb
 require 'rails_helper'
 
 RSpec.describe Presentation, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+  subject { FactoryGirl.create(:presentation) }
+
+  describe 'valid model' do
+    it { is_expected.to be_valid }
```
graygilmore: https://github.com/thoughtbot/shoulda-matchers

Now seems like a fantastic time to add `shoulda_matchers` to our project so that we can optimize some tests like this:

```ruby
it { is_expected.to validate_presence_of :thing }
```

app/controllers/presentations_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      redirect_to @presentation
+    else
+      render 'new'
+    end
+  end
+
+  private
+
+  def presentation_params
+    # binding.pry
```
graygilmore: Whoops!

graygilmore (changes requested):
I'd love for these commit messages to see some updates. Ideally we want something that looks like:

```
The thing that we did

Here is a brief description, taking up any amount of sentences and paragraphs, that explain why this change was necessary and any additional caveats that a future developer might want to know.
```

spec/models/presentation_spec.rb
gregdaynes: Yay! I wasn't sure if we should add it in or not.

graygilmore: I'd add it at as the first commit of the PR.

app/controllers/presentations_controller.rb
gregdaynes: What if future devs need to use pry, but forget how to do it?!

I wish comments had a facepalm emoji

app/models/presentation.rb
gregdaynes: right! I originally had it part of the view, then moved it to the controller, and settled on the model. Forgot about writing a test for it. Thanks!

app/views/presentations/_form.html.erb
```diff
@@ -0,0 +1,35 @@
...
+  </p>
+
+  <p>
+    <%= f.label :duration %><br>
+    <%= f.text_field :duration %> Seconds
+  </p>
+
+  <p>
+    <%= f.label :event %><br>
+    <%= f.select :event_id, Event.all.collect { |e| [ e.name, e.id ] }, { :include_blank => '-- select event --' } %>
```
qr8r: Instead of using collect here, this would be a good situation to use the AR `:pluck` method. 
You might also want to consider using the `collection_select` rails helper

also `:include_blank => '-- select event --'` is old hash syntax

qr8r (changes requested):

app/models/presentation.rb
```diff
@@ -1,4 +1,17 @@
...
+    if person
+      person.first_name
+    else
+      presenter
+    end
```
jacquesporveau: Seems like a good place for a ternary statement.

app/views/presentations/_form.html.erb
```diff
@@ -0,0 +1,35 @@
...
+  </p>
+
+  <p>
+    <%= f.label :presenter %><br>
+    <%= f.select :person_id, Person.all.collect { |p| [ p.first_name, p.id ] }, { :include_blank => '-- select presenter --' } %>
```
jacquesporveau: Old hash here too.

app/views/presentations/new.html.erb
```diff
@@ -0,0 +1,4 @@
+<h1>New Presentation</h1>
+
+<%= render 'form' %>
+
```
jacquesporveau: Whitespace

spec/factories.rb
```diff
@@ -3,19 +3,20 @@
...
   end
 
   factory :presentation do
-    title 'Test Presentation'
-    content 'Lorem ipsum dolor'
-    date_of_presentation Date.new(2017, 2, 18)
+    topic 'Being Awesome'
+    description 'How to be awesome and stuff'
+    duration 900
+
```
jacquesporveau: Whitespace.

gregdaynes: was intentonal to separate out associated factories

spec/models/presentation_spec.rb
```diff
@@ -1,5 +1,154 @@
...
+                                   presenter: 'test',
+                                   person_id: person.id)
+      end
+
+      it { is_expected.to have_attributes(person: person, presenter: 'test') }
+      it { is_expected.to_not be_valid }
+    end
+  end
+
+  describe 'topic' do
```
qr8r: Double testing the shoulda matchers,

app/controllers/presentations_controller.rb
```diff
@@ -0,0 +1,32 @@
+class PresentationsController < ApplicationController
```
brendandeere: Yikes this is a big commit....

qr8r (dismissed):

spec/models/presentation_spec.rb
gregdaynes: Yeah... I did this because in a previous pull request for event (i think) Gray commented on my tests using context, but only having one side of it in that describe, but having the other side in the validate. In these cases I can remove the context all together.

app/views/presentations/_errors.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if @presentation.errors.any? %>
+  <div class="error-explanation">
+    <h2 class="error-explanation__header">
```
graygilmore: Are we using these classes anywhere?

app/controllers/presentations_controller.rb
```diff
@@ -6,4 +6,27 @@ def show
...
+      redirect_to @presentation
+    else
+      render 'new'
+    end
+  end
+
+  private
+
+  def presentation_params
+    params.require(:presentation)
```
graygilmore: I would love to normalize our indentation a little bit. Unfortunately we're not quite sure how to enforce this with Rubocop yet but we don't to write this as:

```ruby
def presentation_params
   params.require(:presentation).permit(
    :presenter,
    :topic,
    :duration,
    :description,
    :person_id,
    :event_id
  )
 end
```

spec/controllers/presentations_controller_spec.rb
```diff
@@ -15,4 +15,39 @@
...
   end
+
+  describe 'POST #create' do
+    subject { post :create, params: { presentation: params } }
+    let!(:person) { FactoryGirl.create(:person) }
+    let!(:event) { FactoryGirl.create(:event) }
+
+    context 'valid params' do
+      let(:params) do
+        FactoryGirl.attributes_for(:presentation,
```
graygilmore: As mentioned above, would love to see:

```ruby
FactoryGirl.attributes_for(
  :presentation,
  thing: thing
)
```

graygilmore (changes requested):

app/views/presentations/_errors.html.erb
gregdaynes: Nope, will remove. Legacy stuff from when I was starting the project. Since I've been recycling the _error partial I created then.

spec/models/presentation_spec.rb
```diff
@@ -1,5 +1,89 @@
...
+
+    it { is_expected.to belong_to :person }
+    it { is_expected.to belong_to :event }
+
+    it { is_expected.to respond_to :presenter_name }
+
+    it { is_expected.to be_valid }
+  end
+
+  describe 'person and presenter' do
```
graygilmore: This looks like it's describing validations.

spec/models/presentation_spec.rb
```diff
@@ -1,5 +1,89 @@
...
+  describe 'validate model' do
+    subject { FactoryGirl.create(:presentation) }
+
+    it { is_expected.to validate_presence_of :topic }
+    it { is_expected.to validate_presence_of :duration }
+
+    it { is_expected.to belong_to :person }
+    it { is_expected.to belong_to :event }
+
+    it { is_expected.to respond_to :presenter_name }
```
graygilmore: But not `person_or_presenter`? I don't think the `respond_to` adds value here as we'll be testing the methods below anyways. Instead of adding another check for `person_or_presenter` I'd just remove this altogether.

spec/controllers/presentations_controller_spec.rb
```diff
@@ -7,4 +7,12 @@
     it { is_expected.to render_template :new }
     it { is_expected.to be_successful }
   end
+
+  describe 'GET #show' do
+    subject { get :show, params: { id: 1 } }
```
graygilmore: There is a chance (albeit small) that the presentation being created here might not have an ID of `1`. You could write this like:

```
describe "GET #show" do
  let(:presentation) { FactoryGirl.create :presentation }
  subject { get :show, params:  { id: presentation.id } }

  # stuff
end
```

Using the presentation in the subject will also allow us to removing the bang from the let.

graygilmore (changes requested):
Just a couple more!

graygilmore (approved):

app/models/presentation.rb
```diff
@@ -1,4 +1,24 @@
...
+  validate  :person_or_presenter
+
+  def presenter_name
+    person ? person.first_name : presenter
+  end
+
+  private
+
+  def person_or_presenter
+    return unless person.present? && presenter.present?
```
jacquesporveau, line 20: I always found `unless` to read a little funny. Worth mentioning that you could also do
```
return if person.nil? && presenter.nil?
```

spec/controllers/presentations_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  describe 'GET #new' do
+    subject { get :new }
+
+    it { is_expected.to render_template :new }
+    it { is_expected.to be_successful }
+  end
+
+  describe 'GET #show' do
+    subject { get :show, params: { id: 1 } }
+    let!(:presentation) { FactoryGirl.create(:presentation) }
```
jacquesporveau: Instead of sending in `params: { id: 1 }` I think it would be better to send in `params: { id: presentation.id }`

spec/models/presentation_spec.rb
```diff
@@ -1,5 +1,87 @@
...
+        let(:presentation) do
+          FactoryGirl.attributes_for(
+            :presentation,
+            event_id: event.id,
+            presenter: 'test'
+          )
+        end
+
+        it { is_expected.to be_valid }
+        it { is_expected.to have_attributes(person: nil, presenter: 'test') }
```
jacquesporveau: I would rather see `presenter: presentation.presenter`

spec/models/presentation_spec.rb
```diff
@@ -1,5 +1,87 @@
...
+        let(:presentation) do
+          FactoryGirl.attributes_for(
+            :presentation,
+            event_id: event.id,
+            presenter: 'test',
+            person_id: person.id
+          )
+        end
+
+        it { is_expected.to have_attributes(person: person, presenter: 'test') }
```
jacquesporveau: Same thing here

spec/models/presentation_spec.rb
```diff
@@ -1,5 +1,87 @@
...
+        it { is_expected.to_not be_valid }
+      end
+    end
+  end
+
+  describe 'presenter_name' do
+    context 'when presenter is not present' do
+      subject { FactoryGirl.create(:presentation) }
+
+      it { is_expected.to have_attributes(presenter_name: 'John') }
```
jacquesporveau, line 86: This is kind of a weird test.

When a presenter is not present a presentation has a presenter name of John? Shouldn't it be nil or something?

app/models/presentation.rb
gregdaynes: Agreed, but it keeps it consistent with the other validators above for person/presenter

spec/models/presentation_spec.rb
gregdaynes: its testing the dynamic property when person is set, and presenter is nil. In the Factories, this is specified as 'John'

PR #20, opened 2017-02-24
Add Seed Data

gregdaynes:
Seed data for interacting with the application on development.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,45 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+# Person Seeder
+5.times do |i|
+  Person.create(
+    first_name: 'Person',
+    last_name: "#{i}",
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -5,3 +5,59 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
```
graygilmore: While we're in here we might as well nuke these comments.

db/seeds.rb
```diff
@@ -5,3 +5,59 @@
...
+    first_name: 'Person',
+    last_name: "#{i}",
+    email: "person.#{i}@example.com",
+    date_of_birth: Date.new(1970,1,1),
+    password: 'abc123',
+    password_confirmation: 'abc123',
+    gender: 'Seed'
+  )
+
+  people.push(person)
```
graygilmore: I don't think we need to be story these in an array. Later on when you want to select a random user you could do something like:

```ruby
person_id: Person.pluck(:id).sample
```

This logic can apply to all of the seeds.

graygilmore (changes requested):

db/seeds.rb
gregdaynes: .sample is amazing!

qr8r (approved):

graygilmore (approved):

PR #21, opened 2017-02-24
WIP add user roles

natyork:


db/migrate/20170224005515_create_people_roles.rb
```diff
@@ -0,0 +1,8 @@
+class CreatePeopleRoles < ActiveRecord::Migration[5.0]
+  def change
+    create_table :people_roles do |t|
+      t.belongs_to :person, index: true
```
qr8r, line 4: We also probably want to add foreign key contraints

db/migrate/20170224003932_create_roles.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRoles < ActiveRecord::Migration[5.0]
+  def change
+    create_table :roles do |t|
+      t.string :name
+
+      t.timestamps
```
qr8r, line 6: These should be in separate commits with their respective db/schema changes

db/migrate/20170224005515_create_people_roles.rb
```diff
@@ -3,6 +3,9 @@ def change
     create_table :people_roles do |t|
       t.belongs_to :person, index: true
       t.belongs_to :role, index: true
+
+      add_foreign_key :people, :people_roles
+      add_foreign_key :roles, :people_roles
```
qr8r: This should be squashed into the commit where the create_people_roles migration is introduced

app/controllers/roles_controller.rb
```diff
@@ -0,0 +1,2 @@
+class RolesController < ApplicationController
```
qr8r, line 1: Avoid gerenators for now they do too much behind the scenes. Do it manually so every piece of  code you add is intentional for now

app/models/person_role.rb
```diff
@@ -0,0 +1,7 @@
+class PersonRole < ApplicationRecord
+  belongs_to :people
+  belongs_to :roles
+
+  #validations
```
qr8r: Avoid committing development comments

app/models/role.rb
```diff
@@ -0,0 +1,8 @@
+class Role < ApplicationRecord
+  has_many :people_roles
+  has_many :people, through: :people_roles
+
+  #validations
```
qr8r: Avoid committing development comments

app/models/role.rb
```diff
@@ -0,0 +1,8 @@
+class Role < ApplicationRecord
```
qr8r, line 1: Should be in its own commit along with a role_spec file

app/models/person_role.rb
```diff
@@ -0,0 +1,7 @@
+class PersonRole < ApplicationRecord
```
qr8r, line 1: Should be in its own commit along with a person_role spec file

app/controllers/roles_controller.rb
qr8r: Also, should be its own commit along with the spec file

app/models/person.rb
```diff
@@ -2,6 +2,9 @@ class Person < ApplicationRecord
   has_many :person_presentations
   has_many :presentations, through: :person_presentations
 
+  has_many :people_roles
+  has_many :roles, through: :people_roles
```
qr8r: Add shoulda matcher specs for these associations

db/schema.rb
```diff
@@ -32,6 +32,13 @@
     t.index ["email"], name: "index_people_on_email", unique: true
   end
 
+  create_table "people_roles", force: :cascade do |t|
```
qr8r: Should be split and included with the relevant migrations

spec/models/role_spec.rb
```diff
@@ -0,0 +1,6 @@
+# spec/models/role_spec.rb
+require 'rails_helper'
+
+RSpec.describe Role, type: :model do
```
qr8r: Should be squashed into the commit that introduces roles

spec/controllers/roles_controller_spec.rb
```diff
@@ -1,5 +1,46 @@
 require 'rails_helper'
 
 RSpec.describe RolesController, type: :controller do
+  describe "GET #index" do
```
qr8r: Avoid committing empty pending specs.

Pending specs can be useful, but they must still implement an expectation. That way RSpec will yell at you if the spec suddenly passes

qr8r (changes requested):
Really good start to this feature, however there is lots we can do to improve this. 

1. A commit should be the smallest piece of functional, cohesive code you can group together. Functional meaning that the change does not break the project (even if it doesnt entirely work yet) and coherent meaning that the change makes sense to be grouped together.

   An example of a commit that is too large in this PR is the commit that introduces the PeopleRoles model, the Role model and the RoleController. This can be broken into smaller chunks without becoming inhoherent. 

    An example of incoherent commits in this PR are the commits that introduce migrations without the associated migration. If you try and run the code at this point, rails will raise a warning about missing migrations. The code will work after migrations are run but there will be local changes that arent committed. While technically functional, the schema changes clearly belong with this commit for it to truely work.
  (in some environments we dont actually run migrations, but rather just load the schema directly and if the schema chages aren't present, they would not be picked up) 

    There were no commits that werent functional in this PR, good job! 

2. Prs should be the smallest grouping of (functional and coherent) commits that implement a whole feature.
    While a commit might introduce functional code that doesnt accomplish a feature, a PR must implement a feature in its entirety.
    As an example, it would be reasonable to have a commit that defined the method: 
```ruby
# Implement adding API
# This method accepts two numbers and returns another number
def add_two_numbers(n1, n2)
  5 
end
```
A PR would have to include commits that fully implemented and tested the behaviour of the `add_two numbers method`

3. We use git as a tool to communicate with other developers. Every commit and every PR should have a well written message/description indicating what behaviour has changes, why that change is needed, and any other information that might be relevant to someone looking at a commit in the future.

app/models/role.rb
```diff
@@ -0,0 +1,13 @@
...
+  has_many :person_roles
+  has_many :people, through: :person_roles
+
+  before_save do
+    name.downcase!
+  end
```
jacquesporveau: Could probably one line this `before_save { name.downcase! }`

spec/models/role_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  describe 'validations' do
+    it { is_expected. to validate_presence_of :name }
+    it { is_expected. to validate_uniqueness_of(:name).case_insensitive }
+    it { is_expected. to be_valid }
+
+    it { is_expected. to have_many :person_roles }
+    it { is_expected. to have_many(:people).through(:person_roles) }
+
+    it "expects name to be stored lower-case" do
```
jacquesporveau: Don't want to be a penny picker but just a heads up there is some inconsistency between single and double quotes being used in here. Generally you only use double quotes when you are performing string interpolation inside the string.

jacquesporveau (changes requested):

jacquesporveau (approved):

jacquesporveau:
Don't need to make my changes. They are just suggestions.

spec/models/role_spec.rb
natyork: Thanks! That's good to know :)

natyork:
I've updated changes from Brendan & Dennis's reviews. Requesting another review!

spec/models/role_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    it { is_expected. to validate_presence_of :name }
+    it { is_expected. to validate_uniqueness_of(:name).case_insensitive }
+    it { is_expected. to be_valid }
+
+    it { is_expected. to have_many :person_roles }
+    it { is_expected. to have_many(:people).through(:person_roles) }
+
+    it 'expects name to be stored lower-case' do
+      subject.name = "EVeNt Planner"
+      subject.save
```
jacquesporveau, line 16: Worth mentioning that these two lines are the same as
```
subject.update(name: 'EVeNt Planner')
```

natyork: Beauty! Thanks Dennis.

graygilmore: This is very close! For readability it might be nicer to write this spec like this:

```ruby
require "rails_helper"

RSpec.describe Role, type: :model do
  it { is_expected.to validate_presence_of :name }
  it { is_expected.to validate_uniqueness_of(:name).case_insensitive }
  it { is_expected.to be_valid }

  it { is_expected.to have_many :person_roles }
  it { is_expected.to have_many(:people).through(:person_roles) }

  context "when the name is mixed-case" do
    let(:role) { FactoryGirl.build(:role, name: "MIxED CaSe") }

    it "saves the name in lower case" do
      role.save!
      expect(role.name).to eq "mixed case"
    end
  end
end
```

graygilmore (changes requested):

jacquesporveau:
I am going to be taking over this issue from @natyork. I am closing this PR as the work I am doing will be different than this.

PR #22, opened 2017-02-24
Presentation creation

gregdaynes:


app/views/presentations/_form.html.erb
```diff
@@ -23,7 +23,7 @@
 
   <p>
     <%= f.label :presenter %><br>
-    <%= collection_select(:presentation, :person_id, Event.all, :id, :name, include_blank: true) %>
+    <%= collection_select(:presentation, :person_id, Person.all, :id, :first_name, include_blank: true) %>
```
graygilmore, line 26: First name is bound to bite us at some point but we can cross that bridge later :wink:

graygilmore (approved):

app/views/presentations/_form.html.erb
gregdaynes: Oh I know. I regret it already. Just wait for the seed data.

PR #23, opened 2017-02-25
Feature/google oauth

jacquesporveau:
- Users need to be able to log in to the staff tracker app via their google authenticated account.
 - If the User does not already have an account with staff tracker then we want to create a new account for them and sign them in on the spot.
 - Users need to be able to log out.


jacquesporveau:
Jenkins test this please

jacquesporveau:
Jenkings test this please

app/models/person.rb
```diff
@@ -33,4 +27,27 @@ class Person < ApplicationRecord
             if: -> { new_record? || !password_confirmation.nil? }
 
   has_secure_password
+
+  scope :from_omniauth, ->(auth) do
+    where(provider: auth.provider, uid: auth.uid)
+  end
```
natyork: This is neat! I didn't know this scope method existed, that is so useful!

db/schema.rb
```diff
@@ -29,6 +29,10 @@
     t.datetime "created_at",      null: false
     t.datetime "updated_at",      null: false
     t.string   "password_digest"
```
natyork, line 31: If we are using Google OAuth, do we still need to store a password digest?

app/models/person.rb
```diff
@@ -33,4 +27,27 @@ class Person < ApplicationRecord
             if: -> { new_record? || !password_confirmation.nil? }
 
   has_secure_password
+
```
natyork: Do we still need passwords if we are using OAuth?

app/models/person.rb
```diff
@@ -33,4 +27,27 @@ class Person < ApplicationRecord
...
+    self.first_name = auth.extra.raw_info.given_name
+    self.last_name = auth.extra.raw_info.family_name
+    self.email = auth.extra.raw_info.email
+    self.provider = auth.provider
+    self.uid = auth.uid
+    self.oauth_token = auth.info.name
+    self.oauth_expires_at = Time.zone.at(auth.credentials.expires_at)
+    self.password = 'NZdBrEYjGacEERsW8b5w'
+    self.password_confirmation = 'NZdBrEYjGacEERsW8b5w'
+    save!
```
natyork: Can we remove password from our model if using OAuth?

natyork (commented):
Only one real question from me, and that is whether we want to keep a password digest in our model if we are using OAuth - I've tried to highlight where I've caught references to passwords in this series of commits, though we may want to do a separate story to remove from app if that makes more sense.

app/models/person.rb
jacquesporveau: :taco:

jacquesporveau:
To reply to @natyork here. I am not sure what route we are taking whether we are exclusively going be using oauth or we want to keep this password field for another sign in method yet to be implemented. Perhaps @brendandeere or @graygilmore could weigh in?

app/models/person.rb
```diff
@@ -33,4 +27,27 @@ class Person < ApplicationRecord
...
+
+  def save_from_omniauth(auth)
+    self.first_name = auth.extra.raw_info.given_name
+    self.last_name = auth.extra.raw_info.family_name
+    self.email = auth.extra.raw_info.email
+    self.provider = auth.provider
+    self.uid = auth.uid
+    self.oauth_token = auth.info.name
+    self.oauth_expires_at = Time.zone.at(auth.credentials.expires_at)
+    self.password = 'NZdBrEYjGacEERsW8b5w'
```
qr8r: This should be set to some value using SecureRandom, or RandomHex

app/models/person.rb
```diff
@@ -33,4 +27,27 @@ class Person < ApplicationRecord
...
+    where(provider: auth.provider, uid: auth.uid)
+  end
+
+  def self.omniauth_retrieval_or_save(auth)
+    from_omniauth(auth).first_or_initialize.tap do |user|
+      user.save_from_omniauth(auth)
+    end
+  end
+
+  def save_from_omniauth(auth)
```
qr8r: This method doesnt belong here. This should be in 'auth/:provider/callback' controller action. It doesnt have to redirect to the same sessions#create action as when a user signs up from our site, instead of with google auth

app/models/person.rb
```diff
@@ -33,4 +27,27 @@ class Person < ApplicationRecord
             if: -> { new_record? || !password_confirmation.nil? }
 
   has_secure_password
+
+  scope :from_omniauth, ->(auth) do
+    where(provider: auth.provider, uid: auth.uid)
+  end
+
+  def self.omniauth_retrieval_or_save(auth)
```
qr8r: I think this should be handled inthe Oauth callback

spec/controllers/sessions_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      post :create
+      expect(session[:person_id]).not_to be_nil
+    end
+  end
+
+  describe "DELETE #destroy" do
+    before { post :create }
+    it 'clears the session' do
+      expect(session[:person_id]).not_to be_nil
+      delete :destroy
```
qr8r: Should be a subject

spec/controllers/sessions_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it 'creates a new session' do
+      post :create
+      expect(session[:person_id]).not_to be_nil
+    end
+  end
+
+  describe "DELETE #destroy" do
+    before { post :create }
+    it 'clears the session' do
+      expect(session[:person_id]).not_to be_nil
```
qr8r: Can be a `expect.to change`

spec/spec_helper.rb
```diff
@@ -110,3 +110,27 @@
   Kernel.srand config.seed
 =end
 end
+ENV["RAILS_ENV"] ||= 'test'
+require File.expand_path("../../config/environment", __FILE__)
+require 'rspec/rails'
+require 'rspec/autorun'
+
+OmniAuth.config.test_mode = true
+omniauth_hash = {
```
qr8r: This doesnt belong in the spec helper. I would create a support directory and perform this setup in a file there.

It can then be required in the rails helper

qr8r (changes requested):

spec/spec_helper.rb
```diff
@@ -91,22 +88,26 @@
...
   Kernel.srand config.seed
 =end
 end
+ENV["RAILS_ENV"] ||= 'test'
+require File.expand_path("../../config/environment", __FILE__)
+require 'rspec/rails'
+require 'rspec/autorun'
+
+# Include all support files
+Dir[Rails.root.join("spec/support/**/*.rb")].each { |file| require file }
```
qr8r, line 113: :+1:

qr8r (approved):
:confetti_ball:

jacquesporveau:
It's a celebration :1st_place_medal:

PR #24, opened 2017-02-27
Feature/upcoming events

gregdaynes:
Feature adds route /upcoming-events to list all present and future events.

[Ticket](https://www.pivotaltracker.com/n/projects/1976733)

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -8,6 +8,10 @@ class EventsController < ApplicationController
     @event = Event.new
   end
 
+  def upcoming
+    @events = Event.upcoming
+  end
+
   def create
     @event = Event.new(event_params)
 
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -8,4 +8,6 @@ class Event < ApplicationRecord
             presence: true
   validates :name,
             presence: true
+
+  scope :upcoming, -> { where('date >= ?', Time.zone.today) }
 end
--- a/app/views/events/upcoming.html.erb
+++ b/app/views/events/upcoming.html.erb
@@ -1,0 +1,30 @@
+<h1>Upcoming Events</h1>
+
+<% @events.each do |event| %>
+  <div>
+    <h2><%= link_to event.name, event_path(event) %></h2>
+
+    <ul>
+      <li>Location: <%= event.location %></li>
+      <li>Date: <%= event.date %></li>
+      <% if event.presentations.present? %>
+        <li>Presentations
+          <ul>
+            <% event.presentations.each do |presentation| %>
+              <li>
+                <dl>
+                  <dt>Topic</dt>
+                  <dd><%= presentation.topic %></dd>
+
+                  <dt>Presenter</dt>
+                  <dd><%= presentation.presenter_name %></dd>
+                </dl>
+              </li>
+            <% end %>
+          </ul> 
+        </li>
+      <% end %>
+    </ul>
+
+  </div>
+<% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
   resources :people
-  resources :events, only: [:show, :create, :new]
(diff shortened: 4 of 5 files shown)


app/views/events/upcoming.html.erb
```diff
@@ -0,0 +1,32 @@
+<h1>Upcoming Events</h1>
+
+<% @events.each do |event| %>
+  <div>
+    <h2><%= link_to event.name, event_path(event) %></h2>
+
+    <dl>
```
graygilmore: Not the best use case of a `<dl>` here. I'd rather see this be an unordered list.

app/views/events/upcoming.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <dt>Location</dt>
+      <dd><%= event.location %></dd>
+
+      <dt>Date</dt>
+      <dd><%= event.date %></dd>
+
+      <dt>Presentations</dt>
+      <dd>
```
graygilmore: We might want a condition here if there are no presentations on an event.

config/routes.rb
```diff
@@ -3,5 +3,7 @@
   resources :events, only: [:show, :create, :new]
   resources :presentations, only: [:show, :create, :new]
 
+  get '/upcoming-events', to: 'events#upcoming'
```
graygilmore: It might be nice to put this under the events resource:

```ruby
resources :events, only: :blah do
  get "/upcoming-events", on: :collection
end
```

(don't quote me on the exact implementation)

app/controllers/events_controller.rb
```diff
@@ -8,6 +8,11 @@ def new
     @event = Event.new
   end
 
+  def upcoming
+    @events = Event.upcoming
+    render 'upcoming'
```
graygilmore: Is this render method necessary?

graygilmore (changes requested):

config/routes.rb
gregdaynes:
```
resources :events, only: [:show, :create, :new, :upcoming] do   
  get "/upcoming-events", to: 'events#upcoming', on: :collection
end                                                             
```

app/models/event.rb
```diff
@@ -8,4 +8,6 @@ class Event < ApplicationRecord
             presence: true
   validates :name,
             presence: true
+
+  scope :upcoming, -> { where('date >= ?', Time.zone.today) }
```
graygilmore, line 12: I think we want `Time.current` here. @brendandeere?

config/routes.rb
```diff
@@ -1,7 +1,11 @@
 Rails.application.routes.draw do
   resources :people
-  resources :events, only: [:show, :create, :new]
+  resources :events, only: [:show, :create, :new, :upcoming] do
+    get "/upcoming-events", to: 'events#upcoming', on: :collection
+  end
   resources :presentations, only: [:show, :create, :new]
 
+  # get '/upcoming-events', to: 'events#upcoming'
```
graygilmore: Whoops.

app/views/events/upcoming.html.erb
```diff
@@ -1 +1,30 @@
...
+
+<% @events.each do |event| %>
+  <div>
+    <h2><%= link_to event.name, event_path(event) %></h2>
+
+    <ul>
+      <li>Location: <%= event.location %></li>
+      <li>Date: <%= event.date %></li>
+      <% if event.presentations.present? %>
+      <li>Presentations
```
graygilmore: Looks like this needs an indent.

graygilmore (changes requested):
Two additional changes:

1. Would love to see "Add 'upcoming' scope to events" as the first line in the first commit message.
2. The last two commits can probably just be squashed together – what do you think? Any reason you kept them separate?

app/models/event.rb
gregdaynes: Because our events don't have times, if the event is for today, it will be skipped with Time.current because it is equivalent to 00:00:00

graygilmore (approved):

PR #25, opened 2017-03-01
Fix /upcoming-events route.

jacquesporveau:
Reference: https://www.pivotaltracker.com/story/show/140277481/comments/164547387

The upcoming-events route had an error where visiting /events/upcoming-events
was sending you to the events#upcoming action instead of the intended
/upcoming-events route corresponding to that action. This change fixes that issue

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,7 +1,7 @@
 Rails.application.routes.draw do
   resources :people
-  resources :events, only: [:show, :create, :new, :upcoming] do
-    get "/upcoming-events", to: 'events#upcoming', on: :collection
+  resources :events, only: [:show, :create, :new] do
+    get 'upcoming', to: 'events#upcoming', on: :collection
   end
   resources :presentations, only: [:show, :create, :new]
 


config/routes.rb
```diff
@@ -1,9 +1,8 @@
 Rails.application.routes.draw do
   resources :people
-  resources :events, only: [:show, :create, :new, :upcoming] do
-    get "/upcoming-events", to: 'events#upcoming', on: :collection
-  end
+  resources :events, only: [:show, :create, :new, :upcoming]
   resources :presentations, only: [:show, :create, :new]
 
   root 'welcome#index'
+  get "/upcoming-events", to: 'events#upcoming'
```
graygilmore: Do you think it's worth it to have the URL be `/events/upcoming` so that we can have a nicely nested route?

jacquesporveau: For me purely from an aesthetic point of view I like it the way it is. The shorter a URL is for me the better but I think as the application grows it might be nice to have it nested as our config file could become a mess and it might just be a bit easier to find for future devs this way. I am like 65% nested and 35% the way it is.

Writing out that comment I think I am going to change it.

jacquesporveau: To further add there are going to be multiple /events routes already so backtracking to /upcoming-events would be weird.

graygilmore: Let's keep it then!

graygilmore:
Whoops, trigger happy. I think I mean "let's change it to `upcoming/events` 😄

jacquesporveau:
Just for clarification Gray and I have talked over Slack and decided to go with `/events/upcoming` which is how the PR stands now.

qr8r (approved):

config/routes.rb
```diff
@@ -1,7 +1,7 @@
 Rails.application.routes.draw do
   resources :people
   resources :events, only: [:show, :create, :new, :upcoming] do
```
qr8r: Might also want to remove `:upcoming` from here. ... not 100% sure

graygilmore: Yup!

graygilmore (changes requested):

graygilmore (approved):

PR #28, opened 2017-03-02
Add presentation validation

jacquesporveau:
Presentations should not be more than 15min in an effort to keep talks concise.

I've made some changes to the FactoryGirl presentation Factory so that the presentation
objects it's pumping out pass validation. I also fixed a spacing error.

### DIFF
--- a/app/models/presentation.rb
+++ b/app/models/presentation.rb
@@ -3,12 +3,14 @@ class Presentation < ApplicationRecord
   belongs_to :event
 
   validates :topic,     presence: true
-  validates :duration,  presence: true, numericality: true
   validates :person,    presence: true, unless: :presenter
   validates :presenter, presence: true, unless: :person
   validates :event,     presence: true
+  validates :duration,  presence: true, numericality: {
+    greater_than: 0, less_than_or_equal_to: 15
+  }
 
-  validate  :person_or_presenter
+  validate :person_or_presenter
 
   def presenter_name
     person ? person.first_name : presenter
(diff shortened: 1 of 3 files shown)


app/models/presentation.rb
```diff
@@ -3,12 +3,14 @@ class Presentation < ApplicationRecord
   belongs_to :event
 
   validates :topic,     presence: true
-  validates :duration,  presence: true, numericality: true
   validates :person,    presence: true, unless: :presenter
   validates :presenter, presence: true, unless: :person
   validates :event,     presence: true
+  validates :duration,  presence: true, numericality: {
```
graygilmore, line 9: What do you think about providing a custom error message to make it a little more clear why we might throw this error?

jacquesporveau: Current error messages being displayed in the form flash include:

Duration must be less than or equal to 15
Duration must be greater than 0

I feel pretty good about these. The application isn't particularly user friendly at this point and my thoughts kind of lead me to think if we are going to set a more friendly message here we should do it for all the validation errors. 

Ultimately I think it's a good idea but with the state the application is in right now I don't think it's the right time.

spec/models/presentation_spec.rb
```diff
@@ -10,11 +10,15 @@
 
     it { is_expected.to validate_presence_of :topic }
     it { is_expected.to validate_presence_of :duration }
-    it { is_expected.to validate_numericality_of :duration }
 
     it { is_expected.to belong_to :person }
     it { is_expected.to belong_to :event }
 
+    it do
```
qr8r: If youre going to use the multiline it statement you should probably add a text description and use an explicit subject. 

it might just be a convention thing... but its a convention that I like a lot

qr8r (changes requested):

graygilmore (approved):

qr8r (approved):

PR #29, opened 2017-03-06
Feature/timebox events

jacquesporveau:
This branch is dependent and set merge into #30. When that PR is merged I will update the branch to merge into master. Do not merge this until #30 is merged.

UPDATE: #30 has been merged so I've updated the branch this is set to merge into to master.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -14,4 +14,8 @@ class Event < ApplicationRecord
   scope :requiring_notification, -> do
     where('date = ?', Date.current - NOTIFICATION_PERIOD)
   end
+
+  def duration
+    presentations.map(&:duration).sum
+  end
 end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,6 +1,7 @@
 # app/models/event.rb
 class Event < ApplicationRecord
   NOTIFICATION_PERIOD = 1.day
+  MAXIMUM_DURATION = Figaro.env.maximum_event_duration.to_i
   has_many :presentations
 
   validates :date,
--- a/spec/models/event_spec.rb
+++ b/spec/models/event_spec.rb
@@ -82,4 +82,33 @@ RSpec.describe Event, type: :model do
       end
     end
   end
+
+  describe '.duration' do
+    subject(:event_duration) { event.duration }
+    context 'when there are no presentations assigned' do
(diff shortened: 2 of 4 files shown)


jacquesporveau:
Tests will fail until Jenkin env file is updated.

spec/models/event_spec.rb
```diff
@@ -82,4 +82,31 @@
       end
     end
   end
+
+  describe '.duration' do
+    subject(:event_duration) { event.reload.duration }
```
qr8r: Why not build the event with the presentations you want, like we did in the last PR instead of building the Event first and then attaching a presentation to it after the fact .

Another fun fact is that if you can get you spec to work with the `build` or `build_stubbed` factory girl method you can really speed up your test suite since you aren't having to create new records in the DB

app/models/event.rb
```diff
@@ -14,4 +14,8 @@ class Event < ApplicationRecord
   scope :requiring_notification, -> do
     where('date = ?', Date.current - NOTIFICATION_PERIOD)
   end
+
+  def duration
+    presentations.map(&:duration).sum
```
qr8r, line 20: No need to change this, just pointing out that this could also be performed at the db level with `#pluck` There are tradeoffs to both perfoming the calculation on the ruby side and doing on the db side. 

Theres no need to change this. Just wanted to point out a fun fact

qr8r (changes requested):
Really small changes here. Good job on the spec. Looks good!

app/models/event.rb
jacquesporveau: Cool I'll have a look to see which one I prefer.

spec/models/event_spec.rb
jacquesporveau: Didn't know that we could be using build this entire time. Nice one! I've changed this file over and will begin using this in all my new work in the future.

app/models/event.rb
jacquesporveau: Alright so we actually can't do this on the DB level because there are instances in which we will want to call this method before presentations have been saved to the database and are hanging around in "memory?" Not sure if that's the right term but the map is a better solution here.

qr8r: Right! Totally reasonable. Thanks for looking into that :)

qr8r (approved):

PR #30, opened 2017-03-06
Refactor spec to work as intended

jacquesporveau:
This spec was passing but not quite doing what we intended it to. I've made some
refactors to bring it back to where it should be.

Replaced some .length calls with static integers just so we can be a little bit
more clear with what exactly is going on here.

### DIFF
--- a/spec/jobs/event_notification_job_spec.rb
+++ b/spec/jobs/event_notification_job_spec.rb
@@ -2,20 +2,30 @@ require 'rails_helper'
 
 describe EventNotificationJob do
   subject(:performed_job) { described_class.perform(event) }
-  let(:event) { FactoryGirl.create(:event, :requiring_notification) }
-  before { FactoryGirl.create(:presentation, event: event) }
 
   describe '.perform' do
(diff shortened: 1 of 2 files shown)


spec/jobs/event_notification_job_spec.rb
```diff
@@ -3,19 +3,20 @@
 describe EventNotificationJob do
   subject(:performed_job) { described_class.perform(event) }
   let(:event) { FactoryGirl.create(:event, :requiring_notification) }
-  before { FactoryGirl.create(:presentation, event: event) }
 
   describe '.perform' do
     context 'when there are multiple presentations for an event' do
-      before { FactoryGirl.create(:presentation, event: event) }
+      before do
+        FactoryGirl.create(:presentation, event: event)
```
qr8r: How about: 

```ruby
FactoryGirl.create(:event, presentations: FactoryGirl.build_list(:presentation, 1))
```

qr8r (changes requested):

qr8r (approved):

PR #31, opened 2017-03-07
Feature/user roles

jacquesporveau:
In this PR we are adding the ability to distinguish between person roles to delegate privileges.
The architecture that I've taken is something like this:

A person can have multiple roles. Admin is implemented in this PR but there could be a 'blogger' roll etc. introduced in the future.

Using the pundit gem we mark which controller actions a person can visit. By default we are making all data altering actions available to admins only.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -1,5 +1,7 @@
 class Person < ApplicationRecord
   has_many :presentations
+  has_many :person_roles
+  has_many :roles, through: :person_roles
 
   before_save do
     email.downcase!
@@ -22,4 +24,8 @@ class Person < ApplicationRecord
             uniqueness: { case_sensitive: false }
 
   has_secure_password
+
+  def admin?
+    roles.detect(&:admin?).present?
+  end
 end
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -1,0 +1,6 @@
+class Role < ApplicationRecord
+  has_many :person_roles
+  has_many :people, through: :person_roles
+
+  enum name: { admin: 0 }
+end
--- a/db/migrate/20170306230433_create_roles.rb
+++ b/db/migrate/20170306230433_create_roles.rb
@@ -1,0 +1,9 @@
+class CreateRoles < ActiveRecord::Migration[5.0]
+  def change
+    create_table :roles do |t|
+      t.column :name, :integer
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20170306230913_create_person_roles.rb
+++ b/db/migrate/20170306230913_create_person_roles.rb
@@ -1,0 +1,10 @@
+class CreatePersonRoles < ActiveRecord::Migration[5.0]
+  def change
+    create_table :person_roles do |t|
+      t.references :person, index: true
+      t.references :role, index: true
+    end
+    add_foreign_key :person_roles, :people
+    add_foreign_key :person_roles, :roles
+  end
+end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -8,6 +8,10 @@ FactoryGirl.define do
     password_confirmation 'abc123'
     date_of_birth Date.new(1970, 1, 1)
     gender 'Computer'
+
+    trait :admin do
+      roles { build_list(:role, 1, :admin) }
+    end
   end
 
   factory :event do
@@ -33,6 +37,12 @@ FactoryGirl.define do
 
     trait :requiring_notification do
       event { FactoryGirl.build(:event, :requiring_notification) }
+    end
+  end
+
+  factory :role do
+    trait :admin do
+      name :admin
     end
   end
 end
--- a/spec/models/person_spec.rb
+++ b/spec/models/person_spec.rb
@@ -9,6 +9,7 @@ RSpec.describe Person, type: :model do
     it { is_expected.to validate_uniqueness_of(:email).case_insensitive }
     it { is_expected.to have_secure_password }
     it { is_expected.to be_valid }
+    it { is_expected.to have_many(:roles) }
 
     it "expects first_name to be longer than 1 character" do
       subject.first_name = "A"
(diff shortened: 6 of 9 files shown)


jacquesporveau:
Just gonna jot this down for memories sake when I start tomorrow. Will delete when done.

Write helper method on Person model called admin?

roles.include(:person) etc.

Once done that head over to ze docks.
https://github.com/elabs/pundit

db/migrate/20170306230913_create_person_roles.rb
```diff
@@ -0,0 +1,8 @@
+class CreatePersonRoles < ActiveRecord::Migration[5.0]
+  def change
+    create_table :person_roles do |t|
+      t.belongs_to :person, index: true
```
qr8r: Can you add foreign key contraints here please?

db/migrate/20170306230433_create_roles.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRoles < ActiveRecord::Migration[5.0]
+  def change
+    create_table :roles do |t|
+      t.string :name
```
qr8r: I might have split this into two commits called: 

Add Roles
Add UserRoles

Also, string is a fair choice here. But I think this could also be a really good place for and `enum`
http://api.rubyonrails.org/classes/ActiveRecord/Enum.html

app/models/person.rb
```diff
@@ -1,5 +1,7 @@
 class Person < ApplicationRecord
   has_many :presentations
+  has_many :person_roles
+  has_many :roles, through: :person_roles
```
qr8r, line 4: Same, I might have split adding these 2 models into separate commits.

app/models/person.rb
```diff
@@ -24,4 +24,8 @@ class Person < ApplicationRecord
             uniqueness: { case_sensitive: false }
 
   has_secure_password
+
+  def admin?
```
qr8r, line 28: This is where an enum might be super cool

spec/factories.rb
```diff
@@ -8,6 +8,10 @@
     password_confirmation 'abc123'
     date_of_birth Date.new(1970, 1, 1)
     gender 'Computer'
+
+    trait :admin do
+      roles { [FactoryGirl.create(:role, :admin)] }
```
qr8r: :+1: on using the trait. :-1: on jumping straight to creating the role.

This means this will always try and create a new role, even if the programmer tried to `build` a person .
I would prefer `roles { build_list(...) } `

spec/models/person_spec.rb
```diff
@@ -27,4 +27,20 @@
       expect(subject.gender).to eq(subject.gender.downcase)
     end
   end
+
+  describe 'admin?' do
```
qr8r, line 31: really nice spec! possibly another edge case would be if the user had more than 1 role. But this is Still a very well written spec

Gemfile
```diff
@@ -37,12 +37,15 @@ gem 'bcrypt', '~> 3.1.7'
 
 gem 'bootstrap', '~> 4.0.0.alpha6'
 
-# Use Google oauth for authorization
+# Use Google oauth for google API authorization
```
qr8r: I might have implemented Pundit as a separate PR from adding roles. 
Even though they are related, they are kind of separate features

app/policies/application_policy.rb
```diff
@@ -0,0 +1,28 @@
+class ApplicationPolicy
+  attr_reader :person, :record
+
+  def initialize(person, record)
+    @person = person
+    @record = record
+  end
+
+  def create?
+    @person.admin?
```
qr8r: Since you are defining readers for these instance variables you might want to use them here instead of accessing the instance variable directly

qr8r (changes requested):
I forgot to publish this review on wednesday :(

db/migrate/20170306230913_create_person_roles.rb
jacquesporveau: @brendandeere to clarify you are looking for a unique constraint on it?

app/models/person.rb
jacquesporveau: Done

app/models/person.rb
jacquesporveau: Done.

spec/factories.rb
jacquesporveau: Split this up into two PRs and decided with the new enum style we are going to handle this differently so just dropped the entire commit.

spec/models/person_spec.rb
jacquesporveau: Thanks but we don't need it anymore haha! With enum there is a built in rails method that we don't need to test.

app/policies/application_policy.rb
jacquesporveau: I've split this up into two PRs and will address this issue in that resulting split.

spec/models/person_spec.rb
jacquesporveau: Thanks.

Gemfile
jacquesporveau: I've done that now.

db/migrate/20170306230913_create_person_roles.rb
jacquesporveau: I think I've added them in here. After searching I found an article describing how to add foreign keys on a DB level and this it what it mentioned. Kind of strange though because the schema file didn't change between commits so I feel like I could still be missing the mark.

app/models/role.rb
```diff
@@ -0,0 +1,6 @@
+class Role < ApplicationRecord
+  has_many :person_roles
+  has_many :people, through: :person_roles
+
+  enum name: [:admin]
```
natyork: This enum method is cool - I had not seen this used before!

natyork (commented):
Nice work! I don't have any changes for you :)

app/models/role.rb
jacquesporveau: Yeah it's got some cool extra functionalities.

qr8r: Enums are amazing! 

I will caution always using the Hash syntax for them: 
```ruby 
enum name: { 
  admin: 0
}
```
Why do you think that is "safer" than using the array syntax? 

Enums also do amazing things like automatically adding scopes and predicate methods for all items in the enum.

So you can now call `Role.admin`, and get a list of all of the roles that have the admin roles. You can also call `role.admin?` to find out if a role is a particular role.

The above methods might be useful for the Person#admin? method

qr8r (changes requested):
1 tiny enum change

app/models/role.rb
jacquesporveau: I think the reason it is safer is that the order of the array could change over time and by clearly stating that the admin is at the zero index and always has been we avoid some of the dangers of future devs messing up our code.

qr8r (approved):

PR #32, opened 2017-03-09
Feature/pundit autho

jacquesporveau:
Dependent on #31 

Admin will be the only ppl who can modify records. Using the pundit gem we enforce these restrictions

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,8 +1,11 @@
 class ApplicationController < ActionController::Base
+  include Pundit
   protect_from_forgery with: :exception
   helper_method :current_person
 
   def current_person
     @current_person ||= Person.find(session[:person_id]) if session[:person_id]
   end
+
+  alias_method :current_user, :current_person
 end
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -6,6 +6,7 @@ class EventsController < ApplicationController
 
   def new
     @event = Event.new
+    authorize @event
   end
 
   def upcoming
@@ -14,6 +15,7 @@ class EventsController < ApplicationController
 
   def create
     @event = Event.new(event_params)
+    authorize @event
 
     if @event.save
       flash[:notice] = 'Event was successfully created.'
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -5,10 +5,12 @@ class PresentationsController < ApplicationController
 
   def new
     @presentation = Presentation.new
+    authorize @presentation
   end
 
   def create
     @presentation = Presentation.new(presentation_params)
+    authorize @presentation
 
     if @presentation.save
       flash[:notice] = 'Presentation was successfully created'
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -1,0 +1,28 @@
+class ApplicationPolicy
+  attr_reader :person, :record
+
+  def initialize(person, record)
+    @person = person
+    @record = record
+  end
+
+  def create?
+    person.admin?
+  end
+
+  def new?
+    create?
+  end
+
+  def update?
+    person.admin?
+  end
+
+  def edit?
+    update?
+  end
+
+  def destroy?
+    person.admin?
+  end
+end
--- a/spec/controllers/events_controller_spec.rb
+++ b/spec/controllers/events_controller_spec.rb
@@ -1,34 +1,54 @@
 require 'rails_helper'
 
 RSpec.describe EventsController, type: :controller do
-  describe 'POST #create' do
-    subject { post :create, params: { event: params } }
+  context 'the current person is an admin' do
+    let(:admin_person) { FactoryGirl.build(:person, :admin) }
 
-    context 'with valid attributes' do
-      let(:params) { FactoryGirl.attributes_for(:event) }
+    before { mock_pundit_user_as(admin_person) }
 
-      it 'redirects' do
-        expect(subject.status).to eq(302)
+    describe 'POST #create' do
+      subject { post :create, params: { event: params } }
+
+      context 'with valid attributes' do
+        let(:params) { FactoryGirl.attributes_for(:event) }
+
+        it 'redirects' do
+          expect(subject.status).to eq(302)
+        end
+        it 'renders the event' do
+          expect(subject).to redirect_to(Event.last)
+        end
+        it 'changes the event count +1' do
+          expect { subject }.to change(Event, :count).by(1)
+        end
       end
-      it 'renders the event' do
-        expect(subject).to redirect_to(Event.last)
-      end
-      it 'changes the event count +1' do
-        expect { subject }.to change(Event, :count).by(1)
+
+      context 'with invalid attributes' do
+        let(:params) { FactoryGirl.attributes_for(:event, name: nil) }
+
+        it 'responds 200' do
+          expect(subject.status).to eq(200)
+        end
+        it 'does not change event count' do
+          expect { subject }.to_not change(Event, :count)
+        end
+        it 'redirects to :new' do
+          expect(subject).to render_template(:new)
+        end
       end
     end
+  end
+  context 'the current person is a non-admin' do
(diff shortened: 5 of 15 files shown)


app/policies/application_policy.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def edit?
+    update?
+  end
+
+  def destroy?
+    @person.admin?
+  end
+end
```
jacquesporveau: From @brendandeere 

Since you are defining readers for these instance variables you might want to use them here instead of accessing the instance variable directly

spec/policies/event_policy_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe EventPolicy, type: :model do
+  describe EventPolicy do
+    it 'inherits from ApplicationPolicy' do
+      expect(described_class).to be < ApplicationPolicy
```
natyork, line 6: This is cool - I wouldn't have thought to test the inheritance!

spec/controllers/events_controller_spec.rb
```diff
@@ -1,6 +1,13 @@
 require 'rails_helper'
 
 RSpec.describe EventsController, type: :controller do
+  let(:admin_person) { FactoryGirl.build(:person, :admin) }
+
+  before do
+    allow_any_instance_of(EventsController).to receive(:pundit_user)
+      .and_return(admin_person)
+  end
```
natyork: I recall you mentioning that you added the before block in to ensure that the user you created in factory girl had the correct authorizations for the spec. Is there a way in rspec to describe your before block such that it reads a little more semantically?

spec/controllers/presentations_controller_spec.rb
```diff
@@ -1,6 +1,13 @@
 require 'rails_helper'
 
 RSpec.describe PresentationsController, type: :controller do
+  let(:admin_person) { FactoryGirl.build(:person, :admin) }
+
+  before do
+    allow_any_instance_of(PresentationsController).to receive(:pundit_user)
+      .and_return(admin_person)
+  end
```
natyork: Same as previous comment for event controller spec.

natyork (commented):
I really like how simple the pundit implementation is. I don't have any proposed changes, but I do have a question about describing before blocks in rspec (see inline comments).

app/controllers/application_controller.rb
```diff
@@ -1,8 +1,11 @@
 class ApplicationController < ActionController::Base
+  include Pundit
   protect_from_forgery with: :exception
   helper_method :current_person
 
   def current_person
     @current_person ||= Person.find(session[:person_id]) if session[:person_id]
   end
+
+  alias pundit_user current_person
```
graygilmore: Do we want to refer to this person as a `pundit_user`? Might be nicer to keep the gem's language out of our code.

app/controllers/events_controller.rb
```diff
@@ -14,6 +15,7 @@ def upcoming
 
   def create
     @event = Event.new(event_params)
+    authorize @event
```
graygilmore, line 18: Might be nice to do a:

```ruby
before_filter :authorize, only: [:new, :create]
```

As I imagine we'll be authorizing more things in the future.

app/controllers/presentations_controller.rb
```diff
@@ -5,10 +5,12 @@ def show
 
   def new
     @presentation = Presentation.new
+    authorize @presentation
```
graygilmore, line 8: Some `before_filter` comment here.

spec/controllers/events_controller_spec.rb
```diff
@@ -1,6 +1,13 @@
 require 'rails_helper'
 
 RSpec.describe EventsController, type: :controller do
+  let(:admin_person) { FactoryGirl.build(:person, :admin) }
```
graygilmore: Would love to see us add a context for the case where the user is *not* an admin.

spec/policies/application_policy_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    policy_methods.each do |method|
+      describe method.to_s do
+        it 'returns true' do
+          expect(subject.send(method)).to eq(true)
+        end
+      end
+    end
+  end
+
+  context 'the person is a pleb' do
```
graygilmore: As it's currently a public project we might want to use different language 😄

spec/controllers/presentations_controller_spec.rb
```diff
@@ -1,6 +1,13 @@
 require 'rails_helper'
 
 RSpec.describe PresentationsController, type: :controller do
+  let(:admin_person) { FactoryGirl.build(:person, :admin) }
+
+  before do
+    allow_any_instance_of(PresentationsController).to receive(:pundit_user)
```
graygilmore: Something about how this is written doesn't make me feel confident that we're testing this code properly. Do we need to stub this?

graygilmore (dismissed):

app/controllers/application_controller.rb
jacquesporveau: We have to I believe. The gem searches for a `current_user`. If you are using a different naming  convention (`current_person`) the docs say you should just alias it this way.

spec/controllers/presentations_controller_spec.rb
jacquesporveau: Yeah because in a controller spec there is no session or concept of a `current_person` so we stub this to return a user that has the priveleges we are trying to test for. If this line code wasn't here the controller would just hit the unauthorized and error out.

spec/policies/application_policy_spec.rb
jacquesporveau: :laughing:

app/controllers/presentations_controller.rb
jacquesporveau: I'll try to find a nice way to implement this. As I recall a before filter is called before controller method is called.  Since we need to have the instance available to authorize this it might be kind of tricky but I'll dive on in.

spec/controllers/events_controller_spec.rb
jacquesporveau: I think the only way I could do that would be to write a comment explaining my actions. Gray also didn't know why this was here so that might be a good idea.

app/controllers/presentations_controller.rb
jacquesporveau: Looks like we can do an `around_filter`. We would have to throw yield statements in place of all the authorize statements so I don't think it really cleans any thing up. Not really feeling that approach so much.

app/controllers/application_controller.rb
qr8r: Could we alias current_user to be current_person here? 

Would that work?

spec/policies/application_policy_spec.rb
```diff
@@ -21,9 +21,9 @@
     end
   end
 
-  context 'the person is a pleb' do
-    subject { ApplicationPolicy.new(pleb, 'any object') }
-    let(:pleb) { FactoryGirl.build(:person) }
```
qr8r: Could technically be squashed in the git history so that correct language is always used in our history

jacquesporveau: MB thought I did that.

app/controllers/application_controller.rb
jacquesporveau: Yeahhhhh we could do that. Feel like it isn't terribly clear when to use current_user and when to use current_person in that case tho. This makes me lean further over to having our person classed renamed to user. I just think somewhere down the road we could end up using current_person and current_user interchangeably and that would kind of suck.

With pundit_user no one gonna be calling that by mistake methinks.

app/controllers/events_controller.rb
qr8r: @graygilmore I think this could be interesting to do. But given we will be loading different resources for different actions, we would need to write something like what Cancan provides for 'intelligently' loading the resource based on the action the controller is performing. 

I think for now being explicit is ok

qr8r (approved):

PR #33, opened 2017-03-13
Feature/homepage

natyork:
https://www.pivotaltracker.com/n/projects/1976733

In this PR a home page has been created that shows the next upcoming event. 

There are two main areas of the layout, the header and the main content area:
  -The header is created as a partial and included in the layout; and
  -The main content area to show the next event, was created in a partial that is rendered from the welcome index.

SCSS partials were created for:
  -global styles
  -header styles
  -main content area styles


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170223213553) do
+ActiveRecord::Schema.define(version: 20170227210710) do
```
graygilmore: I have a feeling this commit should not be in this PR.

app/assets/images/stembolt_logo.svg
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="248px" height="45px" viewBox="0 0 248 45" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 40.3 (33839) - http://www.bohemiancoding.com/sketch -->
```
graygilmore: Here's an awesome tool that will help cut the size of this file down: https://jakearchibald.github.io/svgomg/

app/views/welcome/index.html.erb
```diff
@@ -0,0 +1 @@
+<%= render 'next-event', next_event: @events.first%>
```
graygilmore: In this commit (6f0ebb808cd58793e1e4d7d9824334a4cc4e3579) you've added a partial that doesn't exist yet.

app/assets/stylesheets/_global.scss
```diff
@@ -0,0 +1,24 @@
...
+html {
+  font-size: 62.5%;
+}
+
+body {
+  color: #222222;
+  font-family: sans-serif;
+  font-size: 1.6rem;
+  line-height: 2.5rem;
+  margin: 24px 25px 25px;
```
graygilmore: Prefer to leave the body's padding alone and instead wrap our content in some kind of div.

app/assets/stylesheets/_global.scss
```diff
@@ -0,0 +1,24 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-size: 62.5%;
```
graygilmore: Any reason for this?

app/assets/stylesheets/_global.scss
```diff
@@ -0,0 +1,24 @@
+* {
+  box-sizing: border-box;
```
graygilmore: A slightly nicer way to do this:

```css
html {
  box-sizing: border-box;
}

*,
*:before,
*:after {
  box-sizing: inherit;
}
```
https://css-tricks.com/box-sizing/#article-header-id-6

app/assets/stylesheets/_global.scss
```diff
@@ -0,0 +1,24 @@
...
+
+html {
+  font-size: 62.5%;
+}
+
+body {
+  color: #222222;
+  font-family: sans-serif;
+  font-size: 1.6rem;
+  line-height: 2.5rem;
```
graygilmore: Prefer line-heights to be unitless values so that they always return the value that you expect.

https://github.com/StemboltHQ/information/blob/master/sass_styleguide.markdown

app/assets/stylesheets/_global.scss
```diff
@@ -0,0 +1,24 @@
...
+  margin: 24px 25px 25px;
+}
+
+footer,
+header {
+    display: block;
```
graygilmore: Some indenting issues here.

app/assets/stylesheets/_global.scss
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+a {
+  color: #0F5496;
```
graygilmore: Should we store this color in a Sass variable?

app/assets/stylesheets/application.scss
```diff
@@ -14,4 +14,9 @@
  *= require_self
  */
 
-@import "bootstrap"
+// App
+@import 'header';
+@import 'main';
+
+//Global
+@import 'global';
```
graygilmore: Should global styles be imported before everything else?

app/assets/stylesheets/_header.scss
```diff
@@ -0,0 +1,32 @@
+header {
+  margin: 0 auto;
+  width: 325px;
```
graygilmore: Instead of defining a width could we just leave it off completely?

app/assets/stylesheets/_main.scss
```diff
@@ -0,0 +1,43 @@
+.main {
+  margin: 46px auto 0;
+  width: 325px;
```
graygilmore: What will happen for mobile users that have a width of `320px`? (that's the size of an iPhone 5)

Maybe we could just put a `max-width: 600px` on here instead?

app/assets/stylesheets/_main.scss
```diff
@@ -0,0 +1,43 @@
...
+    margin: 0 0 1.8rem;
+  }
+
+  h2 {
+    font-size: 1.6rem;
+    font-weight: bold;
+    margin: 0;
+  }
+
+  h3 {
```
graygilmore: Do we want all of these styling for `.main` only? Should they be put in a separate `_typography.scss` file?

app/views/welcome/_next-event.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class='main'>
+  <h1> Next Event </h1>
+  <% if !next_event %>
```
graygilmore: It reads a little nicer if we start with the thing that we want as opposed to the fallback:

```ruby
<% if thing_we_want %>

<% else %>

<% end %>
```

graygilmore (changes requested):

app/assets/stylesheets/_global.scss
natyork: Set this value so the root font size would be 10px given a default browser font-size of 16px - made rem calculation easy for other font-sizes. I understood this was a standard approach. Do you use a different approach?

graygilmore: I usually leave the font size here or set it to an explicit value (16px). The `font-size: 62.5%` feels a little bit dated (what if users have a different font size set?) but if you like it feel free to keep it!

app/assets/stylesheets/_global.scss
natyork: That sounds like a great idea! Done!

app/assets/stylesheets/application.scss
natyork: Yes, that makes sense to me :) Updated.

app/assets/images/stembolt_logo.svg
```diff
@@ -0,0 +1,33 @@
...
+                        <path d="M186.091,0.9985 C186.1495,0.8415 186.141,0.696 186.071,0.5855 C185.996,0.479 185.859,0.4205 185.69,0.4205 L172.932,0.4205 C172.762,0.4205 172.627,0.479 172.553,0.5855 C172.478,0.6925 172.4745,0.838 172.533,0.999 L173.293,2.999 C173.432,3.362 173.8795,3.6725 174.2745,3.6725 L176.5845,3.6725 C177.1095,3.6725 177.538,4.0975 177.538,4.626 L177.538,16.836 C177.538,17.211 177.8405,17.5135 178.2155,17.5135 L180.4055,17.5135 C180.778,17.5135 181.0865,17.211 181.0865,16.836 L181.0865,4.625 C181.0865,4.0975 181.5115,3.672 182.034,3.672 L184.366,3.672 C184.756,3.672 185.201,3.362 185.338,2.9985 L186.09,0.9985 L186.091,0.9985 Z"></path>
+                    </g>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
jacquesporveau: We are missing a new line char at the end of this file. I would find a way to configure your .vimrc file to do this on save or something.

app/views/layouts/application.html.erb
```diff
@@ -9,7 +13,7 @@
   </head>
 
   <body>
-    <%= render 'layouts/site-nav' %>
+    <%= render 'layouts/header' %>
```
jacquesporveau, line 16: The header partial is created in the next commit so adding the header in this commit would effectively break functionality. I would add this line in your next commit.

app/views/welcome/_next-event.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <p>There are no upcoming events. <%= link_to 'Create one', new_event_path %>  to get the party started.<p>
```
jacquesporveau: Let's HTTP party!

app/views/welcome/_next-event.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <% else %>
+    <h2><%= next_event.name %></h2>
+    <h3><%= next_event.date.strftime("%B %d, %Y")%></h3>
+    <% if next_event.presentations.empty? %>
+      <p>There are currently no presentations for this event. <%= link_to 'Sign up', new_presentation_path %> to teach some people something!</p>
+    <% else %>
+      <p> Presentations from: </p>
+      <ul>
+        <% next_event.presentations.each do |presentation| %>
+          <li><span><%= presentation.presenter %> &ndash; <%= presentation.topic %> (<%=presentation.duration %> min)</span></li>
```
jacquesporveau: Not sure how rubocop isn't barfing at this but you're over 80 char. I would try splitting this line up into a couple lines.

app/assets/stylesheets/_global.scss
natyork: True enough - I like the root font size at 10px, so it make sense to me to specify 10px explicitly. I've updated accordingly!

app/assets/stylesheets/_main.scss
natyork: Yes that makes sense. Is there a Stembolt standard practice for what belongs in typography vs global?

app/views/welcome/_next-event.html.erb
qr8r: Agreed! You can always flip a negative logical statement to be positive. It is usually easier for humans to understand: 
`if next_event.nill?` than it is to understand ` if !next_event`.

In this you also get to choose which code goes in the `if` and which code goes in the `else`

app/views/welcome/_next-event.html.erb
qr8r: Rubocop only lints `.rb` files. I care less about the 80 char rule in templates... But in this case you might be able to use some indentation to make this a little more readable.

I also think that `<%= presentation.presenter %> &ndash; <%= presentation.topic %> (<%=presentation.duration %> min)` might be a good candidate for interpolation

qr8r (dismissed):

app/assets/stylesheets/_main.scss
graygilmore: Not in particular, just experience. A good rule of thumb to start with: if something doesn't need to be specific right now then leave it as a global rule.

app/assets/stylesheets/_main.scss
natyork: I have updated this as max-width: 325px for the time being. I had a chat with Amanda who suggested the best way to gauge how wide the container should be is to figure out what the widest container for all views will be and style consistently. Since we do not yet have mocks for the other views, she has offered to do some for us :)

natyork:
Changes from @brendandeere, @graygilmore & @jacquesporveau incorporated and pushed up.

app/assets/stylesheets/_global.scss
```diff
@@ -0,0 +1,18 @@
...
+  font-family: sans-serif;
+  font-size: 1.6rem;
+  line-height: 1.6;
+  margin: 0;
+}
+
+footer,
+header {
+  display: block;
+}
```
jacquesporveau: Could the header bit of this CSS be moved into _header.scss?

app/views/welcome/_next-event.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <ul>
+        <% next_event.presentations.each do |presentation| %>
+          <li><%= "#{presentation.presenter} \u2013 #{presentation.topic}
+          (#{presentation.duration} min)" %></li>
+        <% end %>
+      </ul>
+
+    <% else %>
+        <p>There are currently no presentations for this event.
+        <%= link_to 'Sign up', new_presentation_path %>
```
jacquesporveau: Identation.

app/views/welcome/_next-event.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        to teach some people something!</p>
+    <% end %>
+
+  <% else %>
+    <p>There are no upcoming events.
+    <%= link_to 'Create one', new_event_path %>
```
jacquesporveau: Indent.

jacquesporveau (commented):
Super close! Just a couple little things I found.

app/assets/stylesheets/_global.scss
natyork: I pulled applicable css from normalize.css and put it in the global scss file, thinking that was a good place for it but in this case it makes sense that it be included with the relevant partial.

app/views/welcome/_next-event.html.erb
natyork: I think this one is ok actually - there is a nested if statement in this view and this else is part of the outer block.

jacquesporveau: Sorry was referring to the `link_to` tag.

```
<p>There are no upcoming events.
  <%= link_to 'Create one', etc... %>
</p>
```

Your link to tag is nested inside a p tag so we would indent to show the nesting.

natyork:
Changes from @jacquesporveau incorporated and pushed up.

app/views/welcome/_next-event.html.erb
natyork: Rgr, i see it now :)

app/assets/stylesheets/_global.scss
```diff
@@ -0,0 +1,12 @@
+html {
+  box-sizing: border-box;
```
graygilmore, line 2: Whoops, I believe this is set up wrong. This style is correct but we're missing the inheritance.

https://www.paulirish.com/2012/box-sizing-border-box-ftw/

app/assets/stylesheets/_header.scss
```diff
@@ -0,0 +1,34 @@
+header {
+  display: block;
```
graygilmore: A `header` will be `display: block;` by default.

graygilmore (changes requested):

app/assets/stylesheets/_global.scss
natyork: woops, i might have messed that up when i was squashing commits. I will update.

qr8r:
Jenkins test this please

graygilmore (approved):

jacquesporveau:
Worth mentioning that it looks like you've added a `_header.html.erb` file without removing the old `_site-nav.html.erb`. If we are going to be moving forward with the header file then let's delete the old one.

natyork:
@brendandeere Jenkins failing due to test coverage. Dennis and I paired on bringing test coverage up to 100% for master branch this morning, including test for the welcome controller which is missing here. Once RSpec-100 PR is merged, I will rebase this branch and update welcome controller spec as required, then push back up.

natyork:
@jacquesporveau Agreed that we can remove `_site-nav.html.erb` however, I didn't want to nuke this until we sorted out what was going on with the login that you are working on. If you don't need it, I will remove for this PR.

PR #34, opened 2017-03-16
[WIP] - Update presentations policy

jacquesporveau:
We are not going to be limiting presentation creation to
admin users anymore. This commit sets the presentation policy
to reflect that.

I've also removed the admin context from the presentation
controller specs. The reason for this being is there are
no actions in the controller as of now in which an admin
has higher priveleges over a regular user.


graygilmore:
Is this attached to a story?

jacquesporveau:
Yep just throwing this up here more so I can just start looking at the diff.

https://www.pivotaltracker.com/story/show/140356189

@graygilmore

PR #39, opened 2017-03-20
Feature/sign in mandatory

jacquesporveau:
A user should not be able to roam around the site without first signing in. This PR introduces a sign in page, a application controller method to redirect on pages we want to keep hidden from un signed in users, implements that method across the needed controllers and has an slightly updated header.


graygilmore:
👍 for a sign-in page but a user should be able to navigate without logging in. Would love to see those changes made before putting this to review.

PR #40, opened 2017-03-20
Add nil check to policies

jacquesporveau:
There are instances in which a user will not be
signed in whilst still navigiating our app. This nil
check is designed to avoid 500's when that environment is present

Pivotal: https://www.pivotaltracker.com/story/show/141909165


graygilmore (approved):
Makes sense to me.

PR #41, opened 2017-03-20
Feature/edit presentations

natyork:
https://www.pivotaltracker.com/story/show/140277883

This PR updates presentation controller and associated views to allow an admin to edit/update, or delete a presentation.

This PR is dependent on https://github.com/StemboltHQ/staff_tracker/tree/feature/sign-in-mandatory as an admin user needs to be logged in to have update/destroy authorization.

### DIFF
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -12,6 +12,11 @@ class PresentationsController < ApplicationController
     authorize @presentation
   end
 
+  def edit
+    @presentation = Presentation.find(params[:id])
+    authorize @presentation
+  end
+
   def create
     if current_person && current_person.admin?
       @presentation = Presentation.new(admin_presentation_params)
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -13,9 +13,13 @@ class PresentationsController < ApplicationController
   end
 
   def create
-    @presentation = Presentation.new(presentation_params).tap do |presentation|
-      presentation.person = current_person
+    if current_person && current_person.admin?
+      @presentation = Presentation.new(admin_presentation_params)
+    else
+      create_params = presentation_params.merge(person: current_person)
+      @presentation = Presentation.new(create_params)
     end
+
     authorize @presentation
 
     if @presentation.save
@@ -27,6 +31,17 @@ class PresentationsController < ApplicationController
   end
 
   private
+
+  def admin_presentation_params
+    params.require(:presentation).permit(
+      :presenter,
+      :topic,
+      :duration,
+      :description,
+      :event_id,
+      :person_id
+    )
+  end
 
   def presentation_params
     params.require(:presentation).permit(
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -47,6 +47,14 @@ class PresentationsController < ApplicationController
     end
   end
 
+  def destroy
+    @presentation = Presentation.find(params[:id])
+    authorize @presentation
+    @presentation.destroy!
+
+    redirect_to presentations_path
+  end
+
   private
 
   def admin_presentation_params
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -35,6 +35,18 @@ class PresentationsController < ApplicationController
     end
   end
 
+  def update
+    @presentation = Presentation.find(params[:id])
+    authorize @presentation
+
+    if @presentation.update(admin_presentation_params)
+      flash[:notice] = 'Presentation was successfully updated.'
+      redirect_to @presentation
+    else
+      render 'edit'
+    end
+  end
+
   private
 
   def admin_presentation_params
--- a/app/models/presentation.rb
+++ b/app/models/presentation.rb
@@ -26,7 +26,9 @@ class Presentation < ApplicationRecord
   end
 
   def associated_event_has_available_time
-    return if event.nil? || event.duration + duration <= Event::MAXIMUM_DURATION
+    return if duration.nil? ||
+              event.nil? ||
+              event.duration + duration <= Event::MAXIMUM_DURATION
 
     errors.add(
       :event,
--- a/app/views/presentations/edit.html.erb
+++ b/app/views/presentations/edit.html.erb
@@ -1,0 +1,3 @@
+<h1>Edit Presentation</h1>
+
+<%= render 'form', presentation: @presentation %>
--- a/app/views/presentations/new.html.erb
+++ b/app/views/presentations/new.html.erb
@@ -1,3 +1,3 @@
 <h1>New Presentation</h1>
 
-<%= render 'form' %>
+<%= render 'form', presentation: @presentation  %>
--- a/app/views/presentations/show.html.erb
+++ b/app/views/presentations/show.html.erb
@@ -10,3 +10,9 @@
   <dt>Date</dt>
   <dd><%= @presentation.event.date %></dd>
 </dl>
+
+<%= link_to 'Edit', edit_presentation_path(@presentation) %>
+<%= link_to 'Delete', presentation_path(@presentation),
+  method: :delete,
+  data: { confirm: 'Are you sure?' } %>
+
--- a/spec/controllers/presentations_controller_spec.rb
+++ b/spec/controllers/presentations_controller_spec.rb
@@ -1,28 +1,122 @@
 require 'rails_helper'
 
 RSpec.describe PresentationsController, type: :controller do
-  context 'the current person is an admin' do
-    let(:admin_person) { FactoryGirl.build(:person, :admin) }
+  let(:nil_person) { nil }
+  let(:non_admin) { FactoryGirl.build(:person) }
+  let(:admin_person) { FactoryGirl.build(:person, :admin) }
 
-    before { mock_pundit_user_as(admin_person) }
+  describe 'GET #index' do
+    subject { get :index }
 
-    describe 'GET #new' do
-      subject { get :new }
+    context 'the current person is not signed in' do
+      before { mock_pundit_user_as(nil_person) }
+
+      it { is_expected.to render_template :index }
+      it { is_expected.to be_successful }
+    end
+  end
+
+  describe 'GET #show' do
+    subject { get :show, params: { id: presentation.id } }
+    let!(:presentation) { FactoryGirl.create(:presentation) }
+
+    context 'the current person is not signed in' do
+      before { mock_pundit_user_as(nil_person) }
+
+      it { is_expected.to render_template :show }
+      it { is_expected.to be_successful }
+    end
+  end
+
+  describe 'GET #new' do
+    subject { get :new }
+
+    context 'the current person is not signed in' do
+      before { mock_pundit_user_as(nil_person) }
+
+      it 'throws a not authorized error' do
+        expect { subject }.to raise_error(Pundit::NotAuthorizedError)
+      end
+    end
+
+    context 'the current person is a non-admin' do
+      before { mock_pundit_user_as(non_admin) }
 
       it { is_expected.to render_template :new }
       it { is_expected.to be_successful }
     end
+  end
 
-    describe 'POST #create' do
-      subject { post :create, params: { presentation: params } }
-      let!(:person) { FactoryGirl.create(:person) }
-      let!(:event) { FactoryGirl.create(:event) }
+  describe 'GET #edit' do
+    subject { get :edit, params: { id: presentation.id } }
+    let(:presentation) { FactoryGirl.create(:presentation) }
+
+    context 'the current person is a non-admin' do
+      before { mock_pundit_user_as(non_admin) }
+
+      it 'throws a not authorized error' do
+        expect { subject }.to raise_error(Pundit::NotAuthorizedError)
+      end
+    end
+
+    context 'the current person is an admin' do
+      before { mock_pundit_user_as(admin_person) }
+
+      it { is_expected.to render_template :edit }
+      it { is_expected.to be_successful }
+    end
+  end
+
+  describe 'POST #create' do
+    subject { post :create, params: { presentation: params } }
+    let(:event) { FactoryGirl.create(:event) }
+    let(:person) { FactoryGirl.create(:person) }
+
+    context 'the current person is not signed in' do
+      before { mock_pundit_user_as(nil_person) }
+
+      let(:params) do
+        FactoryGirl.attributes_for(
+          :presentation,
+          event_id: event.id
+        )
+      end
+
+      it 'throws a not authorized error' do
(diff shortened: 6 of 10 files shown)


natyork:
Stand-by for review while I fix rubocop offenses.

natyork:
Rubocop offenses fixeed.

jacquesporveau:
Talking with Gray we've decided to move away from a mandatory sign in. I am pretty sure that this PR will solve your problems but perhaps take a look

https://github.com/StemboltHQ/staff_tracker/pull/40

app/views/presentations/edit.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1>Edit Presentation</h1>
+
+<%= render 'form' %>
+
+<%= link_to 'Back', presentations_path %>
```
jacquesporveau: Kind of hesitant to assume that we know where the user is coming from. There is nothing to stop the user from just typing `presentations/1/edit` in their address bar. This back button would not take them to where they came from in that instance.

spec/controllers/presentations_controller_spec.rb
```diff
@@ -90,6 +90,17 @@
         end
       end
     end
+
```
jacquesporveau: Good job on the specs.

app/views/presentations/index.html.erb
```diff
@@ -0,0 +1,29 @@
+<h1>Presentations</h1>
+
+<%= link_to 'New Presentation', new_presentation_path %>
+
+<table>
+  <thead>
+    <tr>
+      <th>Presentation</th>
+      <th>
```
jacquesporveau: Missing closing tag.

natyork:
Added spec for non-admin trying to GET #edit

app/views/presentations/index.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <tbody>
+    <% @presentation.each do |presentation| %>
+      <tr>
+        <td><%= presentation.topic %></td>
+        <td>
+          <%= link_to 'Show', presentation_path(presentation) %>
+          <%= link_to 'Edit', edit_presentation_path(presentation) %>
+          <%= link_to 'Delete', presentation_path(presentation),
+                      method: :delete,
+                      data: { confirm: 'Are you sure?' } %>
```
jacquesporveau: Three links in one table cell seems weird. Interested to see what this looks like.

jacquesporveau (changes requested):

app/views/presentations/edit.html.erb
natyork: Good point! I used the the previously created people index view as a placeholder to give us easier  access to the edit form and delete link. We may want to add a new story to pivotal to format and style desired views.  I have removed the Back link.

app/views/presentations/index.html.erb
natyork: It just puts all three links in one line. It doesn't look really great, and I'm not sure that we should be using tables at all, but as mentioned above, I used the people index view as a template for the time being until we have decided on format, flow, and styling for views.

natyork:
Updated presentations index view per @jacquesporveau comments and rebased commits onto updated master that merged #40

app/controllers/presentations_controller.rb
```diff
@@ -37,6 +37,14 @@ def update
     end
   end
 
+  def destroy
+    @presentation = Presentation.find(params[:id])
+    authorize @presentation
+    @presentation.destroy
```
qr8r: if you are not checking the return value of an AR CRUD method, you should probably use its bang method instead.

`destroy!` will raise an error if the record fails to save.

spec/controllers/presentations_controller_spec.rb
```diff
@@ -90,6 +90,17 @@
...
+    describe 'DELETE #destroy' do
+      let!(:presentation) { FactoryGirl.create(:presentation) }
```
qr8r: doesnt need to be a `let!` you call this method in the subject so the presentation will get initialised then

app/controllers/presentations_controller.rb
```diff
@@ -1,4 +1,8 @@
 class PresentationsController < ApplicationController
+  def index
+    @presentation = Presentation.all
```
qr8r: Not semantic. Presentation implies one record

app/views/presentations/index.html.erb
```diff
@@ -0,0 +1,27 @@
...
+<%= link_to 'New Presentation', new_presentation_path %>
+
+<table>
+  <thead>
+    <tr>
+      <th>Presentation</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @presentation.each do |presentation| %>
```
qr8r: `<%= render @presentations %>`

app/views/presentations/index.html.erb
qr8r: yeah, I'm ok with this. Alternatively we only have links to show and destroy here, and edit is linked to from the show page.

qr8r (changes requested):

app/controllers/presentations_controller.rb
jacquesporveau: Did you mean it raise an error if it fails to destroy?

qr8r: Yes. All of the AR Crud methods have a "bang method" equivalent 

so `create` and `create!` 

The normal method returns true or false, the bang method raises an error. Or convention generally is if you are not checking the output of the method, you should raise an error if it fails 

so: 
```ruby
if thing.create
  stuff
else 
  other stuff
end 

# OR 

thing.create! 
```

natyork:
Updated PR per @brendandeere comments and rebased onto updated master that merged #44. I kept the show/edit/delete links on the presentations#index page but also added edit/delete links to the presentations#show view.

app/controllers/presentations_controller.rb
```diff
@@ -20,6 +29,26 @@ def create
...
+      redirect_to @presentation
+    else
+      render 'edit'
+    end
+  end
+
+  def destroy
+    @presentation = Presentation.find(params[:id])
+    authorize @presentation
+    @presentation.destroy
```
fredericboivin: As Brendan mentioned, this should be `destroy!` since the return value is not used

natyork: Good catch @dangerdogz, I thought I had updated that. Thanks!

spec/controllers/presentations_controller_spec.rb
natyork: @brendandeere The way this test is implemented it doesn't work without the bang because when presentation gets called in the subject block it immediately destroys the presentation, and so the test to check if the presentation count has changed will fail. Is there a better approach overall for this test?

natyork:
Updated  presentations#index view to be consistent with events#index.

natyork:
Merging #43 into master has caused some functionality conflicts WRT an admin user being able to edit the presenter. Will update the controller & form partial accordingly.

app/controllers/presentations_controller.rb
```diff
@@ -9,9 +9,15 @@ def new
   end
 
   def create
-    @presentation = Presentation.new(presentation_params).tap do |presentation|
-      presentation.person = current_person
+    if current_person.admin?
+      @presentation = Presentation.new(admin_presentation_params)
+    else
+      @presentation = Presentation.new(presentation_params)
+                                  .tap do |presentation|
```
qr8r: I dont think #tap is needed here. I believe you can pass a block directly to the #new method, which will yield the new instance.

Alternatively you might be able to the presentation params to add in current_person. 
`Presentation.new(person: current_person, **presentation_params)` 

Or you could also call merge on the presentation params hash 
`Presentation.new(presentation_params.merge(person: current_person))`

spec/controllers/presentations_controller_spec.rb
```diff
@@ -164,6 +198,18 @@
         end
       end
     end
+
+    describe 'PATCH #update' do
```
qr8r: Generally we have 1 describe block for each methods and then nest the contexts below that. Here we have the reverse, The contexts outside the describe blocks

spec/controllers/presentations_controller_spec.rb
qr8r: Right! Given that, you can ignore my previous comment :)

app/views/presentations/new.html.erb
```diff
@@ -1,3 +1,3 @@
 <h1>New Presentation</h1>
 
-<%= render 'form' %>
+<%= render partial: 'form', locals: { presentation: @presentation } %>
```
qr8r: This is older rails syntax. You can pass locals in without the keyword args.

`<%= render 'form', presentation: @presentation %>`

qr8r (changes requested):

app/views/presentations/new.html.erb
jacquesporveau: Nice to know!

app/models/presentation.rb
```diff
@@ -26,6 +26,7 @@ def person_or_presenter
   end
 
   def associated_event_has_available_time
+    return unless duration
```
jacquesporveau: Two things.

1.) We could avoid the `unless` keyword by going `return if duration.nil?` which is a bit more readable

2.) Duration should probably be a required field in the schema. I imagine if someone is creating a presentation we should be ensuring that it has a duration. That is unrelated to this PR and would be feature creep so this is fine but I'll talk it over with @brendandeere and maybe make a resulting issue.

app/views/presentations/show.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= flash[:notice] %>
```
jacquesporveau: Wonder if this should be included in the application layout?

spec/controllers/presentations_controller_spec.rb
```diff
@@ -6,10 +6,19 @@
...
+    describe 'GET #edit' do
+      let(:presentation) { FactoryGirl.create(:presentation) }
+      subject { get :edit, params: { id: presentation } }
 
-      it { is_expected.to render_template :new }
+      it { is_expected.to render_template :edit }
+      it { is_expected.to be_successful }
+    end
+
+    describe 'GET #edit' do
```
jacquesporveau: You're already testing the #edit action above. Dupe test.

natyork:
Please disregard the presentation controller spec for the time being. I messed it up with rebasing so I'm currently fixing.

spec/controllers/presentations_controller_spec.rb
natyork: Cool, that makes sense to me! I spent a while trying to rearrange a lot of these methods to fit into the pattern we already had for the project, but I think it would make it much more organized to do it the way you described! I will refactor and update.

app/models/presentation.rb
natyork: Agree with item 1 and have updated accordingly (another option could be if `errors.any?`, but i like that this is more specific to what this custom validation requires)

WRT item 2, as discussed, there is already a duration validation in the presentation model but it seems that rails will run all validations even if one fails, so if you have a custom validation that is dependent on an attribute being valid, you will have to test explicity for this in your custom validation. Ref http://stackoverflow.com/questions/5966055/controlling-the-order-of-rails-validations

app/views/presentations/show.html.erb
natyork: Great idea!

spec/controllers/presentations_controller_spec.rb
natyork: Fixing up spec after rebase mixup, and will be reorganizing per @brendandeere's comment above.

natyork:
Updated PR per @brendandeere and @jacquesporveau comments. When refactoring the spec for the presentation controller, I rearranged a lot of code, so I did not break the individual changes for each method into the separate commits, as I normally do. This seemed to make the most sense to me in terms of ease of review, but please let me know if you'd prefer to see this broken up into separate commits!

qr8r (approved):

app/views/presentations/show.html.erb
```diff
@@ -10,3 +10,9 @@
   <dt>Date</dt>
   <dd><%= @presentation.event.date %></dd>
 </dl>
+
+<%= link_to 'Edit', edit_presentation_path(@presentation) %>
+<%= link_to 'Delete', presentation_path(@presentation),
+  method: :delete,
+  data: { confirm: 'Are you sure?' } %>
```
jacquesporveau, line 17: I'd kind of like to see these parameters lined up a little more but if rubocop is alright with it then I am too.

spec/controllers/presentations_controller_spec.rb
```diff
@@ -49,45 +143,72 @@
...
+      end
+
+      context 'invalid attributes' do
+        let(:attr) { { topic: nil } }
+
+        it { is_expected.to render_template :edit }
 
-        it 'throws a not authorized error' do
-          expect { subject }.to raise_error(Pundit::NotAuthorizedError)
+        it 'shouldn\'t update presentation' do
```
jacquesporveau: Personally I like to avoid escaping characters in strings but this is fine. 

Some ways you can avoid escaping characters that I've found is just excluding punctuation or using double quotations as the opening and closing string limiters.

spec/controllers/presentations_controller_spec.rb
```diff
@@ -49,45 +143,72 @@
...
         end
       end
     end
+  end
 
-    describe 'GET #index' do
-      subject { get :index }
+  describe 'DELETE #destroy' do
+    subject { delete :destroy, params: { id: presentation.id } }
+    let!(:presentation) { FactoryGirl.create(:presentation) }
```
jacquesporveau, line 194: Do we need a ! here? I feel like when you make the request the presentation would be created before it was sent so it would be available in the controller.

jacquesporveau (approved):
A question and several quality of life ideas. Most of them are personal preference so I am not marking these as needed changes.

app/views/presentations/show.html.erb
natyork: As discussed, rubocop only checks the .rb files so this file wouldn't have been been checked. However you bring up an excellent point - we seem to be using different indenting styles in our views to break up long lines. I just found this thoughtbot styleguide that has an opinion on this:
https://github.com/thoughtbot/guides/blob/master/style/erb/sample.html.erb

spec/controllers/presentations_controller_spec.rb
natyork: As discussed, the way this test is implemented it doesn't work without the bang because when presentation gets called in the subject block it immediately destroys the presentation, and so the test to check if the presentation count has changed will fail.

spec/controllers/presentations_controller_spec.rb
natyork: Fair enough - not really sure where my head was at there ;) Now it's super bugging me, so I'm going to change it!

PR #42, opened 2017-03-20
Update header partial

jacquesporveau:
I've added a sign in/out link to the header

Pivotal: https://www.pivotaltracker.com/story/show/142092931

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -4,6 +4,13 @@
     <ul>
       <li><%= link_to 'Upcoming Events', upcoming_events_path %></li>
       <li><%= link_to 'New Presentation', new_presentation_path %></li>
+      <li>
+        <% if current_person %>
+          <%= link_to 'Sign Out', signout_path %>
+        <% else %>
+          <%= link_to 'Sign In', google_sign_in_path %>
+        <% end %>
+      </li>
     </ul>
   </div>
 </header>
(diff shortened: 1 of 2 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -4,6 +4,13 @@
     <ul>
       <li><%= link_to 'Upcoming Events', upcoming_events_path %></li>
       <li><%= link_to 'New Presentation', new_presentation_path %></li>
+      <li>
+        <% if current_person %>
+          <%= link_to 'Sign Out', signout_path %>
+        <% else %>
+          <%= link_to 'Sign In', '/auth/google_oauth2' %>
```
graygilmore: Is there a defined path for this in our routes?

jacquesporveau: Yep it's `get 'auth/:provider/callback', to: 'authentication#google'`

graygilmore: Let's use the `_path` helper then. A little better than manually specifying the string.

jacquesporveau: I think the gem adds that route in for us or something. It's not defined in our routes but I've tested it and it works / this is old behaviour from the google oauth PR being re introduced.

jacquesporveau:
For clarities sake the path that I've overwritten in the routes file was not being user anywhere.

app/views/layouts/_header.html.erb
jacquesporveau: Sounds good I've updated the PR to include an aliased route.

qr8r (approved):

PR #43, opened 2017-03-20
Feature/presentation signup

jacquesporveau:
Main features being:
 - Presentations can be created by non-admin users that are signed in.
 - Presentations are auto assigned to current user on creation
 - Event dropdown only shows upcoming events

Pivotal: https://www.pivotaltracker.com/n/projects/1976733

### DIFF
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -9,7 +9,9 @@ class PresentationsController < ApplicationController
   end
 
   def create
-    @presentation = Presentation.new(presentation_params)
+    @presentation = Presentation.new(presentation_params).tap do |presentation|
+      presentation.person = current_person
+    end
     authorize @presentation
 
     if @presentation.save
@@ -28,7 +30,6 @@ class PresentationsController < ApplicationController
       :topic,
       :duration,
       :description,
-      :person_id,
       :event_id
     )
   end
--- a/app/views/presentations/_form.html.erb
+++ b/app/views/presentations/_form.html.erb
@@ -18,14 +18,7 @@
 
   <p>
     <%= f.label :event %><br>
-    <%= collection_select(:presentation, :event_id, Event.all, :id, :name, prompt: true) %>
-  </p>
-
-  <p>
-    <%= f.label :presenter %><br>
-    <%= collection_select(:presentation, :person_id, Person.all, :id, :first_name, include_blank: true) %>
-    or
-    <%= f.text_field :presenter, placeholder: 'Enter guest name' %>
+    <%= f.collection_select(:event_id, Event.upcoming, :id, :name, prompt: true) %>
   </p>
 
   <p>
--- a/spec/controllers/presentations_controller_spec.rb
+++ b/spec/controllers/presentations_controller_spec.rb
@@ -55,29 +55,12 @@ RSpec.describe PresentationsController, type: :controller do
 
     before { mock_pundit_user_as(non_admin) }
 
-    describe 'GET #new' do
-      subject { get :new }
-
-      it 'throws a not authorized error' do
-        expect { subject }.to raise_error(Pundit::NotAuthorizedError)
-      end
-    end
-
     describe 'GET #show' do
       subject { get :show, params: { id: presentation.id } }
       let!(:presentation) { FactoryGirl.create(:presentation) }
 
       it { is_expected.to render_template :show }
       it { is_expected.to be_successful }
-    end
-
-    describe 'POST #create' do
-      subject { post :create, params: { presentation: params } }
-      let(:params) { FactoryGirl.attributes_for(:presentation) }
-
-      it 'throws a not authorized error' do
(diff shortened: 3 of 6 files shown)


spec/controllers/presentations_controller_spec.rb
```diff
@@ -55,29 +55,12 @@
 
     before { mock_pundit_user_as(non_admin) }
 
-    describe 'GET #new' do
```
harmonymjb, line 58: It may be a good idea to continue to test these controller endpoints.
Controller specs give you access to a response you can assert against.
e.g. `expect(response.status).to eq(401)` or `expect { subject }.to redirect_to(unauthorized_path)`

app/controllers/presentations_controller.rb
```diff
@@ -9,7 +9,12 @@ def new
   end
 
   def create
-    @presentation = Presentation.new(presentation_params)
+    # because of the presentation policy we know a person
+    # will have to be signed in to view this page.
+    # current_person will never be nil
```
harmonymjb: Which page is `this page`?
Do you think the reader would expect that `current_user` _could_ be `nil`?
What is expected to happen if I POST to `/presentations` without logging in? Can this be tested for in the controller spec?

app/views/presentations/_form.html.erb
```diff
@@ -18,11 +18,9 @@
 
   <p>
     <%= f.label :event %><br>
-    <%= collection_select(:presentation, :event_id, Event.all, :id, :name, prompt: true) %>
+    <%= collection_select(:presentation, :event_id, Event.upcoming, :id, :name, prompt: true) %>
   </p>
 
-  <%= hidden_field_tag(:current_person, current_person.id) %>
```
harmonymjb: This line seems unrelated to this commit.
If this line is not needed, the history can be rebased so that it is never added.

spec/controllers/presentations_controller_spec.rb
jacquesporveau: Yeah the reason I took these tests out was with the presentation pundit policy changing, non-admin and admin users share the same privileges as each other when accessing the "new" and "create" controller actions. So if I were to keep these tests here I would just have to change them to be testing the exact same thing that the admin tests are testing.

Feel like if the policy is tested, which it is. We can be sure that this is working as intended.

app/controllers/presentations_controller.rb
jacquesporveau: Yep that's not very clear. I meant you wouldn't be able to view `/presentations/new` without being logged in first.

`current_user` has the potential to be nil if you made the post request from like a ARC or AJAX request outside the app. In that case you would be greeted with a Pundit not authorized error. 

Which is not tested. I'll write up specs for that.

app/views/presentations/_form.html.erb
```diff
@@ -18,14 +18,7 @@
...
     <%= f.label :event %><br>
-    <%= collection_select(:presentation, :event_id, Event.all, :id, :name, prompt: true) %>
-  </p>
-
-  <p>
-    <%= f.label :presenter %><br>
-    <%= collection_select(:presentation, :person_id, Person.all, :id, :first_name, include_blank: true) %>
-    or
-    <%= f.text_field :presenter, placeholder: 'Enter guest name' %>
+    <%= collection_select(:presentation, :event_id, Event.upcoming, :id, :name, prompt: true) %>
```
dangerdogz: This can be associated with the form object and replaced by `<%= f.collection_select(:event_id, Event.upcoming, :id, :name, prompt: true) %>`

Otherwise, the default value will not be selected from the @presentation object

fredericboivin: This can be associated with the form object and replaced by `<%= f.collection_select(:event_id, Event.upcoming, :id, :name, prompt: true) %>`

Otherwise, the default value will not be selected from the @presentation object

fredericboivin (changes requested):

app/views/presentations/_form.html.erb
jacquesporveau: That sounds like a good idea. Worth mentioning that as there is no edit action implemented yet I don't think there is a situation in which this form would appear that a presentation would already have an associated event already assigned.

However this form will surely be used when the edit functionality is introduced so I am going to include it. Cheers thanks!

fredericboivin (approved):

spec/controllers/presentations_controller_spec.rb
harmonymjb: I think the policy test is good. It tests that the policy returning correctly.
The policy is tested in one context, i.e. `new?` returns `true` is meaningful only if you understand the context of the policy's responsibility, and that you're sure it's correctly applied to the controller.
It might seem intuitive that is `new?` returns `true` on the policy means that the controller is functioning properly, but as the complexity of the application grows, you'll want to test the the policy returns `true` and `false` correctly _and_ that your controller produces the responses (`http_status`, redirect, etc.) that you expect.

Keep in mind what the controller's responsibility is and how it is different than the policy's.

PR #45, opened 2017-03-21
Update header partial

jacquesporveau:
I've added a sign in/out link to the header

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -4,6 +4,13 @@
     <ul>
       <li><%= link_to 'Upcoming Events', upcoming_events_path %></li>
       <li><%= link_to 'New Presentation', new_presentation_path %></li>
+      <li>
+        <% if current_person %>
+          <%= link_to 'Sign Out', signout_path %>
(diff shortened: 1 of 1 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -4,6 +4,13 @@
     <ul>
       <li><%= link_to 'Upcoming Events', upcoming_events_path %></li>
       <li><%= link_to 'New Presentation', new_presentation_path %></li>
+      <li>
+        <% if current_person %>
+          <%= link_to 'Sign Out', signout_path %>
+        <% else %>
+          <%= link_to 'Sign In', '/auth/google_oauth2' %>
```
jacquesporveau, line 11: @graygilmore has expressed some concerns with the url being hardcoded. The omniauth gem doesn't expose a routes when I run `rake routes`. Using a hardcoded url seems to be the adopted best practice as railscasts, richonrails and more tutorials are using this method.

graygilmore (approved):
A shame we can't use a path helper. We'll look into this more in the future!

PR #46, opened 2017-03-21
Feature/edit events

natyork:
https://www.pivotaltracker.com/story/show/140278149

This PR updates events controller and associated views to allow an admin to edit/update, or delete an event.

Upon event deletion, any associated presentations are also deleted. A conditional confirmation message for event deletion warns user that presentations for this event will also be deleted.

For views/actions currently limited to admin, a 500 error is displayed if the user is not a logged in admin.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -13,6 +13,11 @@ class EventsController < ApplicationController
     @events = Event.upcoming
   end
 
+  def edit
+    @event = Event.find(params[:id])
+    authorize @event
+  end
+
   def create
     @event = Event.new(event_params)
     authorize @event
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -30,6 +30,18 @@ class EventsController < ApplicationController
     end
   end
 
+  def update
+    @event = Event.find(params[:id])
+    authorize @event
+
+    if @event.update(event_params)
+      flash[:notice] = 'Event was successfully updated.'
+      redirect_to @event
+    else
+      render 'edit'
+    end
+  end
+
   private
 
   def event_params
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -46,6 +46,14 @@ class EventsController < ApplicationController
     end
   end
 
+  def destroy
+    @event = Event.find(params[:id])
+    authorize @event
+    @event.destroy!
+
+    redirect_to events_path
+  end
+
   private
 
   def event_params
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1,5 +1,9 @@
 # app/controllers/events_controller.rb
 class EventsController < ApplicationController
+  def index
+    @events = Event.all
+  end
+
   def show
     @event = Event.find(params[:id])
   end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -2,7 +2,7 @@
 class Event < ApplicationRecord
   NOTIFICATION_PERIOD = 1.day
   MAXIMUM_DURATION = Figaro.env.maximum_event_duration.to_i
-  has_many :presentations
+  has_many :presentations, dependent: :destroy
 
   validates :date,
             presence: true
--- a/app/views/events/edit.html.erb
+++ b/app/views/events/edit.html.erb
@@ -1,0 +1,3 @@
+<h1>Edit Event</h1>
+
+<%= render 'form' %>
--- a/app/views/events/index.html.erb
+++ b/app/views/events/index.html.erb
@@ -1,0 +1,15 @@
+<h1>Events</h1>
+
+<%= link_to 'New Event', new_event_path %>
+
+<table>
+  <thead>
+    <tr>
+      <th>Event</th>
+      <th>Actions</th>
+    </tr>
+  </thead>
+  <tbody>
+    <%= render @events %>
+  </tbody>
+</table>
--- a/app/views/events/show.html.erb
+++ b/app/views/events/show.html.erb
@@ -1,3 +1,5 @@
+<%= flash[:notice] %>
+
 <h1>Event: <%= @event.name %></h1>
 <dl>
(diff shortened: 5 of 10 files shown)


app/views/events/edit.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1>Edit Event</h1>
+
+<%= render 'form' %>
+
+<%= link_to 'Back', events_path %>
```
jacquesporveau: Two things:

1.) This commit message doesn't give any hints that we are creating a view. I don't think it should be split into another commit because we would have tests broken on this commit but maybe we could mention it in our PR message.

2.) Linking to back. We don't know how the user got here we shouldn't assume they came from `/events/:id`

spec/controllers/events_controller_spec.rb
```diff
@@ -68,6 +98,16 @@
       end
     end
 
+    describe 'PATCH #update' do
+      let(:event) { FactoryGirl.create(:event) }
+      subject { patch :update, params: { id: event.id, event: attr } }
+      let(:attr) { { name: 'New Name' } }
```
jacquesporveau: Super minor but I think it's nice to keep the let statements grouped together. I've kind of been adopting the practice of having the subject above all the let statements just because it's more important than regular variables and we always know where to find it.

spec/controllers/events_controller_spec.rb
```diff
@@ -76,6 +76,17 @@
       end
     end
 
+    describe 'DELETE #destroy' do
+      let!(:event) { FactoryGirl.create(:event) }
```
jacquesporveau: Do we need a bang after this let?

app/models/event.rb
```diff
@@ -2,7 +2,7 @@
 class Event < ApplicationRecord
   NOTIFICATION_PERIOD = 1.day
   MAXIMUM_DURATION = Figaro.env.maximum_event_duration.to_i
-  has_many :presentations
+  has_many :presentations, dependent: :destroy
```
jacquesporveau, line 5: Would like to see a spec showing this works.

app/views/events/index.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1>Events</h1>
+
+<%= link_to 'New Event', new_event_path %>
+
+<table>
+  <thead>
+    <tr>
+      <th>Event</th>
+    </tr>
```
jacquesporveau: The event partial that you are loading has two `<td>` tags. I would expect there to be two `<th>` tags.

app/views/events/show.html.erb
```diff
@@ -9,3 +9,6 @@
   <dt>Location</dt>
   <dd><%= @event.location %></dd>
 </dl>
+
+<%= link_to 'Edit', edit_event_path(@event) %>
+<%= render partial: 'delete-link', locals: { event: @event } %>
```
jacquesporveau, line 16: This commit is out of order. The delete link partial hasn't been created yet.

jacquesporveau (changes requested):

app/views/events/show.html.erb
natyork: The delete link partial was created in the previous commit (Add Delete links to #index).

spec/controllers/events_controller_spec.rb
```diff
@@ -6,6 +6,14 @@
 
     before { mock_pundit_user_as(admin_person) }
 
+    describe 'GET #edit' do
+      let(:event) { FactoryGirl.create(:event) }
+      subject { get :edit, params: { id: event } }
```
harmonymjb: Should this be `event.id`?

spec/controllers/events_controller_spec.rb
```diff
@@ -68,6 +76,15 @@
       end
     end
 
+    describe 'GET #edit' do
+      let(:event) { FactoryGirl.create(:event) }
+      subject { get :edit, params: { id: event } }
```
harmonymjb: Should this be `event.id`?

app/controllers/events_controller.rb
```diff
@@ -30,6 +30,18 @@ def create
...
 
+  def update
+    @event = Event.find(params[:id])
+    authorize @event
+
+    if @event.update(event_params)
+      flash[:notice] = 'Event was successfully updated.'
+      redirect_to @event
+    else
+      render 'edit'
```
harmonymjb, line 45: It might be a good idea to notify the user somehow if the update fails.

app/views/events/show.html.erb
jacquesporveau: Oh weird it was out of order earler :confused:

harmonymjb (changes requested):
Changes look good, just a couple of questions.

Also, I'm seeing `Unrecognized author. If this is you, make sure natalie@freerunningtechnologies.com is associated with your account. Click to add email addresses in your account settings.` next to your name on each commit. You might need to double check your git settings.

fredericboivin (approved):
Looks good to me outside of what Graham mentioned already 👍

spec/controllers/events_controller_spec.rb
natyork: Yes :) Even though this works, I like that using event.id is more explicit.

app/controllers/events_controller.rb
natyork: Totally agree and currently if there are errors in updating/creating events they get rendered via the `app/views/events/_errors.html.erb` from the `app/views/events/_form.html.erb` partial (which in turn is rendered from new & edit views.  That pattern already existed in the project so I carried it forward for the addition of edit/update actions. If you can recommend a better approach for this, I think it would be a good idea to create a new story to update this throughout the project.

natyork: Also I realized that the flash notice on success wasn't actually being rendered anywhere so I added that into events#show view.

natyork:
Updates per @omnistegan comments.

harmonymjb (approved):

jacquesporveau (approved):

PR #47, opened 2017-03-22
Feature/auto show tell

jacquesporveau:


lib/models/event_schedule.rb
```diff
@@ -0,0 +1,15 @@
+class EventSchedule
+  def self.populate_upcoming_show_and_tell!
+    upcoming_dates = (Time.zone.today..Time.zone.today + 4.weeks).to_a
```
luukveenis: You could reduce these 2 lines to:
`(Time.zone.today..Time.zone.today + 4.weeks).select(&:friday?)`

spec/models/event_schedule_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      4.times do
+        dates << Date.new(2017, 1, 6 + counter)
+        counter += 7
+      end
+      dates
+    end
+
+    it 'creates show and tell events' do
+      expect { subject }
+        .to change { Event.where(name: 'Show & Tell').length }
```
luukveenis: This could be more explicit in what it's testing, ie:
`expect { subject }.to change { ... }.from(0).to(4)`

spec/models/event_schedule_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.describe EventSchedule, type: :model do
+  describe '.populate_upcoming_show_and_tell!' do
+    subject { described_class.populate_upcoming_show_and_tell! }
+    before { travel_to Date.new(2017, 1, 1) }
+    after { travel_back }
+
+    let(:fridays) do
```
luukveenis: Usually if you're pushing things into an array you can simplify it to a map, something like:
`(0..3).map { |week| Date.new(2017, 1, 6 + (week * 7)) }`

spec/models/event_schedule_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    it 'creates show and tell events' do
+      expect { subject }
+        .to change { Event.where(name: 'Show & Tell').length }
+    end
+
+    it 'creates show and tell events that take place on Friday' do
+      subject
+      Event.where(name: 'Show & Tell').each do |event|
```
luukveenis: I haven't tested this but I'm pretty sure you can do something like:
`expect(Event.where(name: 'Show & Tell').pluck(:date)).to all(be_friday)`

Docs for the `all` matcher:
https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/all-matcher

The `be_` matcher should work for any predicate method

spec/models/event_schedule_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        expect(fridays).to include(event.date)
+      end
+    end
+
+    it 'doesnt create duplicate events' do
+      subject
+      expect { subject }
+        .not_to change { Event.where(name: 'Show & Tell').length }
```
luukveenis: It doesn't matter too much in this test but usually prefer `count` or even `size` to length since `length` will load the full collection into memory.

This post does a reasonable job of explanaing the differences:
http://mensfeld.pl/2014/09/activerecord-count-vs-length-vs-size-and-what-will-happen-if-you-use-it-the-way-you-shouldnt/

luukveenis (changes requested):
Looks good overall! Just a couple minor comments

lib/models/event_schedule.rb
jacquesporveau: Didn't know you could call `.friday?` on date objects. This is helpful.

spec/models/event_schedule_spec.rb
jacquesporveau: Dang man this recommendations are on point. Learning a lot about methods, ways of doing things that I didn't already know.

spec/models/event_schedule_spec.rb
jacquesporveau: I've taken care of this but this comment won't show up as outdated as the line changed is below the line commented on.

luukveenis (approved):
Nice! looks good

jacquesporveau:
UPDATE:

We are going to move lib/models to app/services

spec/models/event_schedule_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
```
luukveenis: You moved the file to `app/services`, so the spec should probably move to `spec/services` as well

spec/models/event_schedule_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+RSpec.describe EventSchedule, type: :model do
```
luukveenis: This probably doesn't need `type: :model` since it's not a model

luukveenis (changes requested):

luukveenis (approved):
Looks good, thanks for making those changes!

PR #49, opened 2017-03-23
Feature/pagination

jacquesporveau:
https://www.pivotaltracker.com/story/show/140356125

If you are looking to test this out you can run the seed script and then visit `/presentations`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,9 @@ gem 'figaro'
 # Use Pundit for authorization
 gem 'pundit'
 
+# Use Kaminari for pagination
+gem 'kaminari'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platform: :mri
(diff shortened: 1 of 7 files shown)


graygilmore (approved):

harmonymjb (approved):
Changes look good! :+1:

The only thing I felt I should comment on is your final commit message. You left this:
`This is meant to be a bare bones view. In the future I could see us adding a lot more into this but that is outside the feature I am currently working on.`

I would recommend keeping the body section of your commits concise and limited to what is relevant to that commit specifically.
Conversation about the scope of the changes, and future plans for expanding on these changes are already happening in the pivitol stories (and possibly the PR) so it might be best to have these comments in that context rather than the commit message.

jacquesporveau:
Thanks for the feedback on that one @omnistegan. I've taken out the unneeded comment. I guess in the future this would be a good thing to bring up in Pivotal or even in the PR itself. I have a nasty habit of leaving code and git comments like this one.

PR #50, opened 2017-03-23
Feature/inline errors

jacquesporveau:
https://www.pivotaltracker.com/story/show/140356551


natyork (commented):
This looks great to me - glad we are removing those error partials, they seemed very repetitive and easy to lose sight of.

config/initializers/modify_rails_form_builder.rb
```diff
@@ -0,0 +1,14 @@
+class ActionView::Helpers::FormBuilder
+  def error_message_for(field_name)
```
graygilmore, line 2: This is a little bit scary! I would love to avoid requiring an initializer for this. Did you consider switching to something like `simple_form`?

jacquesporveau: I did think about it briefly but I was able to make the code happen without it fairly easily. I feel like for the functionality we need right now this is good but I could see perhaps down the road simple form becoming more appealing.

I also remember my past co-worker ranting how much he didn't like simple form so that also swayed my opinion.

qr8r: I'm with Gray for this one. I'm really not a fan of monkey patching. If it does need to happen you could consider using `class_eval` or  `Module#prepend/extend` to inject behaviour, but that's really only slightly better that opening up the class directly.

I think in this case there is no reason for this not to be a view helper which accepts an object and a field name and renders the errors.

config/initializers/modify_rails_form_builder.rb
```diff
@@ -0,0 +1,14 @@
...
+    if self.object.errors[field_name].present?
+      model_name              = self.object.class.name.downcase
+      target_elem_id          = "#{model_name}_#{field_name}"
+
+      "<div class=\"inline-form-error\" for=\"#{target_elem_id}\">"\
+      "#{self.object.errors[field_name].join(', ')}"\
+      "</div>".html_safe
+    end
+  rescue
```
qr8r, line 11: Blanket rescues are scary ... What errors are you expecting here

qr8r (changes requested):

jacquesporveau:
I am going to scrap this work and re-write it using Simple Form.

PR #51, opened 2017-03-23
Feature/dropdown dates

jacquesporveau:
UPDATE: Going to take Nats advice here. I've split the front end functionality into a module and it's good to go code wise but I am going to clean up my git history a bit tomorrow AM. Don't need to review this yet.

UPDATE: We've decided to go down the presenter route as per Brens advice.

https://www.pivotaltracker.com/story/show/141187279

### DIFF
--- a/app/presenters/base_presenter.rb
+++ b/app/presenters/base_presenter.rb
@@ -1,0 +1,8 @@
+class BasePresenter < SimpleDelegator
+  attr_reader :view
+  def initialize(model, view)
+    @view = view
+    @model = model
+    super(model)
+  end
+end
--- a/app/presenters/event_presenter.rb
+++ b/app/presenters/event_presenter.rb
@@ -1,0 +1,5 @@
+class EventPresenter < BasePresenter
+  def display_name
+    "#{@model.name} - #{@model.date}"
+  end
+end
--- a/app/views/presentations/_form.html.erb
+++ b/app/views/presentations/_form.html.erb
@@ -1,3 +1,5 @@
+<% upcoming_events = Event.upcoming.map { |event| EventPresenter.new(event, self) } %>
+
 <%= form_for presentation do |f| %>
   <%= render 'errors' %>
 
@@ -18,7 +20,7 @@
 
   <p>
     <%= f.label :event %><br>
-    <%= f.collection_select(:event_id, Event.upcoming, :id, :name, prompt: true) %>
+    <%= f.collection_select(:event_id, upcoming_events, :id, :display_name, prompt: true) %>
   </p>
   <% if current_person && current_person.admin? %>
     <p>
--- a/spec/presenters/base_presenter_spec.rb
+++ b/spec/presenters/base_presenter_spec.rb
@@ -1,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe BasePresenter do
+end
--- a/spec/presenters/event_presenter_spec.rb
+++ b/spec/presenters/event_presenter_spec.rb
@@ -1,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe EventPresenter do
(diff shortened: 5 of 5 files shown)


spec/models/event_spec.rb
```diff
@@ -121,4 +121,15 @@
       end
     end
   end
+
+  describe '.name_and_date' do
```
graygilmore: This should be `#name_and_date` to signify that it's an instance method.

http://betterspecs.org/#describe

graygilmore (changes requested):

natyork (commented):
Is it a common practice to add methods to the models to help with front end rendering like this? I don't have a good feel for what is generally accepted in this regard, but at first glance it seems like this might get a bit confusing if we keep adding 'view helper' methods to our model. Thoughts? @graygilmore

graygilmore:
@natyork this could easily go into a view helper as well.

app/helpers/event_front_end.rb
```diff
@@ -0,0 +1,5 @@
+module EventFrontEnd
```
qr8r: http://nithinbekal.com/posts/rails-presenters/

I believe this should be a presenter. 

Helpers can be useful for cleaning up view logic, but rails includes every helper in every view context by default

app/models/event.rb
```diff
@@ -1,5 +1,6 @@
 # app/models/event.rb
 class Event < ApplicationRecord
+  include EventFrontEnd
```
qr8r: http://www.javaworld.com/article/2074564/learn-java/a-look-at-the-composite-design-pattern.html

I think a presenter could be a good candidate for the composite design pattern, if you need this logic to be available to  the model

app/helpers/event_front_end.rb
```diff
@@ -0,0 +1,5 @@
...
+  def name_and_date
```
qr8r: This is definitely what is being returned here... but maybe something like "display_name" might be more representative of what this method is doing. That way if we later decide to add the event organizer name to this values, we wont have a method called `name_and_date` that gives you back: "name and date and organizer"

qr8r (changes requested):

app/helpers/event_front_end.rb
jacquesporveau: Score! This is the definitive answer I was looking for.

jacquesporveau:
Alright I've re done this work in the way that @brendandeere described in his comments.

app/presenters/base_presenter.rb
```diff
@@ -0,0 +1,11 @@
+class BasePresenter < SimpleDelegator
+  def initialize(model, view)
+    @view = view
+    @model = model
+    super(model)
+  end
+
+  def h
```
qr8r: So I know the blog post I linked to called this 'h', but I think a better name for this woudl be `view_context`. this could also be a simple attr_reader, and then there is no reason to write a test for it ;)

config/application.rb
```diff
@@ -11,5 +11,8 @@ class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+    config.after_initialize do |app|
+      app.config.paths.add 'app/presenters'
```
qr8r: http://guides.rubyonrails.org/autoloading_and_reloading_constants.html

qr8r (changes requested):
Super minor changes, otherwise :shipit:

jacquesporveau:
@brendandeere I've made your requested changes.

qr8r (approved):

graygilmore (approved):
Fwiw, I'm not a huge fan of presenter classes but this looks pretty reasonable.

PR #52, opened 2017-03-23
Bug/homepage links

natyork:
Conditionally render links on homepage and header that will produce a 500 error if user is not logged in.

Per story rejection comments on:
https://www.pivotaltracker.com/story/show/141187843

### DIFF
--- a/app/views/welcome/_next-event.html.erb
+++ b/app/views/welcome/_next-event.html.erb
@@ -18,16 +18,12 @@
     <% else %>
       <p>
         There are currently no presentations for this event.
-        <%= link_to 'Sign up', new_presentation_path %>
+        <%= render 'shared/new_presentation_link' %>
         to teach some people something!
       </p>
     <% end %>
 
   <% else %>
-    <p>
-      There are no upcoming events.
-      <%= link_to 'Create one', new_event_path %>
-      to get the party started.
-    <p>
+    <p>There are no upcoming events.</p>
   <% end %>
 </div>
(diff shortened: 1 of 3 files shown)


app/views/welcome/_next-event.html.erb
```diff
@@ -18,16 +18,14 @@
     <% else %>
       <p>
         There are currently no presentations for this event.
-        <%= link_to 'Sign up', new_presentation_path %>
+        <%= render 'next-event-links' %>
```
graygilmore: Maybe we could rename this partial to `new_presentation_link` or something like that? That way we could use it in other areas of the system, too.

app/views/welcome/_next-event.html.erb
```diff
@@ -18,16 +18,14 @@
...
         to teach some people something!
       </p>
     <% end %>
 
   <% else %>
     <p>
       There are no upcoming events.
-      <%= link_to 'Create one', new_event_path %>
```
jacquesporveau: With this being removed you can probably have that `<p>` tag be a one line.

app/views/welcome/_next-event.html.erb
jacquesporveau: I think this is fine for now but maybe a hint for the user that they have to sign in to create one in the future. There are currently no presentations for this event... sign in. Feels weird. I like There are currently no presentations for this event... sign in to create one. Either is fine really.

natyork:
Updated per @graygilmore & @jacquesporveau comments

jacquesporveau (commented):
LGTM

qr8r (approved):

PR #53, opened 2017-03-24
Update presentations controller

jacquesporveau:
https://www.pivotaltracker.com/story/show/140356125/comments/166820279

Configured the pagination to show newest presentations at the top

### DIFF
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -1,6 +1,8 @@
 class PresentationsController < ApplicationController
   def index
-    @presentations = Presentation.page(params[:page]).order(:created_at)
+    @presentations = Presentation.joins(:event)
+                                 .merge(Event.order(date: :desc))
+                                 .page(params[:page])
   end
 
   def show


app/controllers/presentations_controller.rb
```diff
@@ -1,6 +1,6 @@
 class PresentationsController < ApplicationController
   def index
-    @presentations = Presentation.page(params[:page]).order(:created_at)
+    @presentations = Presentation.page(params[:page]).order(created_at: :desc)
```
graygilmore: Should it use the `created_at` or the actual date of the presentation? I'd prefer to see them in order of when they were actually presented. What do you think?

jacquesporveau: We had this conversation yesterday as a pod. I feel strongly that I want to see the newest created presentations at the top. For instance there might be a presentation that is being created for way in the future but I want to get excited about it on creation. I want to see new content.

We have a route to do with events called `events/upcoming` I feel like the same route for presentations would be a good place to show upcoming presentations but this being an index page I want to see the newest available content at the top. Makes me feel like there is a reason to be re-visiting this page.

graygilmore: What if I'm backlogging presentations into the system? Should those also appear at the top? `created_at` is useful when the thing you're modelling doesn't have any other timestamps on it. Because our items do have an associated date with them I would expect them to use that for sorting.

I would prefer to see it use the event date for sorting.

jacquesporveau: and so it was...

qr8r (approved):

PR #54, opened 2017-03-24
Bug/homepage presenter name

natyork:
Homepage was only rendering presenter name for guest presenters. Now presenter name is correctly rendered for either guest presenters or for staff tracker users giving presentations.

### DIFF
--- a/app/views/welcome/_next-event.html.erb
+++ b/app/views/welcome/_next-event.html.erb
@@ -10,7 +10,7 @@
 
       <ul>
         <% next_event.presentations.each do |presentation| %>
-          <li><%= "#{presentation.presenter} \u2013 #{presentation.topic}
+          <li><%= "#{presentation.presenter_name} \u2013 #{presentation.topic}
           (#{presentation.duration} min)" %></li>
         <% end %>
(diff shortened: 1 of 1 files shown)


app/views/welcome/_next-event.html.erb
```diff
@@ -10,7 +10,7 @@
 
       <ul>
         <% next_event.presentations.each do |presentation| %>
-          <li><%= "#{presentation.presenter} \u2013 #{presentation.topic}
+          <li><%= "#{presentation.presenter_name} \u2013 #{presentation.topic}
```
jacquesporveau, line 13: Just curious why you are using `\u2013` over `-`?

natyork: So that it displays as an 'em dash' (—) vice an 'en dash' (–) or hypen  (-). The mock given for this story used an em dash, or at least what most resembled an em dash :)

jacquesporveau: Whew nice attention to detail! I wouldn't have noticed.

qr8r (approved):

PR #55, opened 2017-03-24
Update presentation policy

jacquesporveau:
https://www.pivotaltracker.com/story/show/140356259

Ruby target version was changed because Brendan has asked for us to use the safe navigation operator introduced in ruby 2.3

People should be able to edit their own presentations
so they can fix typos etc.

### DIFF
--- a/app/policies/presentation_policy.rb
+++ b/app/policies/presentation_policy.rb
@@ -6,4 +6,12 @@ class PresentationPolicy < ApplicationPolicy
   def new?
     create?
   end
+
+  def update?
+    (person&.admin? || person&.presentations&.include?(record)).present?
+  end
+
+  def edit?
+    update?
+  end
 end
--- a/spec/policies/presentation_policy_spec.rb
+++ b/spec/policies/presentation_policy_spec.rb
@@ -1,9 +1,11 @@
 require 'rails_helper'
 
 RSpec.describe PresentationPolicy, type: :model do
-  policy_methods = [
+  requiring_sign_in_methods = [
     :create?,
-    :new?
+    :new?,
+    :update?,
+    :edit?
   ]
 
   describe PresentationPolicy do
@@ -15,23 +17,69 @@ RSpec.describe PresentationPolicy, type: :model do
       subject { PresentationPolicy.new(nil_person, 'any_object') }
       let(:nil_person) { nil }
 
-      policy_methods.each do |method|
+      requiring_sign_in_methods.each do |method|
         describe method.to_s do
           it 'returns false' do
-            expect(subject.send(method)).to eq(false)
+            expect(subject.send(method)).to be(false)
           end
         end
       end
     end
 
-    context 'the person is a non admin' do
-      subject { PresentationPolicy.new(non_admin, 'any_object') }
-      let(:non_admin) { FactoryGirl.build(:person) }
+    context 'the person is an admin' do
+      subject { described_class.new(admin, presentation) }
+      let(:admin) { FactoryGirl.build(:person, :admin) }
+      let(:presentation) { FactoryGirl.build(:presentation) }
 
-      policy_methods.each do |method|
-        describe method.to_s do
+      describe '.update?' do
+        it 'returns true' do
+          expect(subject.update?).to be(true)
+        end
+      end
+
+      describe '.edit?' do
+        it 'calls the .update? method' do
+          expect(subject).to receive(:update?)
+          subject.edit?
+        end
+      end
+    end
+
+    context 'the person is a non admin' do
+      subject { PresentationPolicy.new(non_admin, presentation) }
+      let(:non_admin) { FactoryGirl.build(:person) }
+      let(:presentation) { FactoryGirl.build(:presentation) }
+
+      describe '.update?' do
+        context 'the presentation being taken action on belongs to the person' do
+          before do
+            non_admin.presentations << presentation
+          end
+
           it 'returns true' do
-            expect(subject.send(method)).to eq(true)
+            expect(subject.update?).to eq(true)
+          end
+        end
+        context 'the presentation being taken action on doesnt belong to the person' do
+          it 'returns false' do
+            expect(subject.update?).to eq(false)
+          end
+        end
+      end
+
+      describe '.edit?' do
+        context 'the presentation being taken action on belongs to the person' do
+          before do
+            non_admin.presentations << presentation
+          end
+
+          it 'returns true' do
+            expect(subject.update?).to eq(true)
+          end
+        end
+        context 'the presentation being taken action on doesnt belong to the person' do
+          it 'returns false' do
+            expect(subject.update?).to eq(false)
           end
         end
       end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,6 +38,7 @@ Metrics/BlockLength:
 Metrics/LineLength:
   IgnoredPatterns:
     - '\w*it.+do'
+    - '\w*context.+do'
 
 Style/Lambda:
     Enabled: false


app/policies/presentation_policy.rb
```diff
@@ -6,4 +6,12 @@ def create?
   def new?
     create?
   end
+
+  def update?
+    person.present? && (person.admin? || person.presentations.include?(record))
```
qr8r: So `.present?` can be a little bit omniscient since it also considers empty string and and empty arrays to be falsy. 
In this case. the person will either be a truthy value or nil. do you really need to check for presence? 

Also ruby 2.3 introduced the guard operator which was discussed in stembolt slack at some point last week

spec/policies/presentation_policy_spec.rb
```diff
@@ -24,16 +26,56 @@
...
+    end
+
     context 'the person is a non admin' do
-      subject { PresentationPolicy.new(non_admin, 'any_object') }
+      subject { PresentationPolicy.new(non_admin, presentation) }
       let(:non_admin) { FactoryGirl.build(:person) }
+      let(:presentation) { FactoryGirl.build(:presentation) }
+
+      describe '.update?' do
+        context %(the presentation being taken action on
```
qr8r: Why the percent litterals instead of just a string?

spec/policies/presentation_policy_spec.rb
```diff
@@ -24,16 +26,56 @@
...
+
+          it 'returns false' do
+            expect(subject.update?).to eq(false)
+          end
+        end
+      end
+
+      describe '.edit?' do
+        it 'calls the .update? method' do
+          expect(subject).to receive(:update?)
```
qr8r: Not really a fan of this because the test requires internal knowlege of how `.edit?` is implemented. I know its done to save test duplication. 

But I would rather explicit tests

spec/policies/presentation_policy_spec.rb
```diff
@@ -24,16 +26,56 @@
...
+
+          before do
+            non_admin.presentations << presentation
+          end
 
-      policy_methods.each do |method|
-        describe method.to_s do
           it 'returns true' do
-            expect(subject.send(method)).to eq(true)
+            expect(subject.update?).to eq(true)
```
qr8r, line 60: https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/be-matchers

qr8r (changes requested):

spec/policies/presentation_policy_spec.rb
jacquesporveau: 80 char

qr8r: Also regular strings can be wrapped on multiple lines, the percent literal really isnt doing anything there that you couldn't do with normal quotes

```
Metrics/LineLength:
  IgnoredPatterns:
    - '\w*it.+do'
    - '\w*context.+do'
```

jacquesporveau: Yeah rubocop likes double line strings as a percent literal. I'll update the rules though.

app/policies/presentation_policy.rb
jacquesporveau: Yeah the reason I put in the `.present?` call is because since this method ends in a ? it is implied that it returns a boolean. If I went `nil && true` I would get back `nil`. The same sort of problem arises with the guard operator.

```
!!(person&.admin? || person&.presentations&.include?(record))
```

Would be the outcome. I would have to swap the `nil` value to a false value. I also can't do a `!.nil?` check because that would error on false values. I've changed it to be the way I listed it above because I understand the importance of not erroring on empty arrays or strings even though I don't think our `current_person` would ever be set to one of those things but I do think that this solution is a little bit cryptic. 

I sent it by Nat this morning and she had a tough time reading it so I feel like if others have trouble reading it then it isn't the best.

jacquesporveau: Alright I've landed on this solution

```
(person&.admin? || person&.presentations&.include?(record)).present?
```

You can't get an empty string or empty array back from and or clause and this avoids the double !

natyork: Wouldn't it be ok if `update?` returned nil since nil is falsey anyway, or does Pundit explode?

At the same time I see that it breaks the convention of the ? at the end of the method name....so i'm just curious.

spec/policies/presentation_policy_spec.rb
jacquesporveau:
```
  expect(obj).to be_truthy  # passes if obj is truthy (not nil or false)
  expect(obj).to be_falsey  # passes if obj is falsy (nil or false)
  expect(obj).to be_nil     # passes if obj is nil
  expect(obj).to be         # passes if obj is truthy (not nil or false)
```

I don't want to check if the value is truthy or falsey. I want to check that it is true or false.
I can do `expect(bleh).to be(true)` but it basically the exact same thing. Another point against choosing `be` or `eq` is it defies the standard we have been using outside this PR.

jacquesporveau:
Jenkins will fail until Ruby version is pushed to > 2.3. We are targeting 2.4 in rubocop so that would be preferred version.

jacquesporveau:
jenkins test this please

jacquesporveau:
jenkins test this please

spec/policies/presentation_policy_spec.rb
qr8r: Thats fair, but as a caution testing explicitly for true or false can bite you.

Logical operators on their own do not guarantee a boolean return value:
`current_user && current_user.orders.last`
```ruby
def has_orders?
  current_user && current_user.order.last
end
```
could return an Order object or nil. Which would fail a eq(true) test

app/policies/presentation_policy.rb
jacquesporveau: This is purely because of the ? convention. The method must be named `.update?` and having a non boolean return value defies those conventions.

spec/policies/presentation_policy_spec.rb
jacquesporveau: Right this kind of relates to the above conversation where I've designed the `.update?` method to only return booleans.

jacquesporveau: Also in the `has_orders?` method. Are they considering it OK that they aren't returning a boolean from that method? I've always thought that ? methods are supposed to return booleans.

qr8r: Predicate methods are intended to have their result only used as if they were truthy or falsy. but the actual value doesnt have to be true or false

app/policies/presentation_policy.rb
```diff
@@ -6,4 +6,12 @@ def create?
   def new?
     create?
   end
+
+  def update?
+    (person&.admin? || person&.presentations&.include?(record)).present?
```
qr8r, line 11: Why not make the initializer a little smarter. 

https://github.com/StemboltHQ/staff_tracker/blob/master/app/policies/application_policy.rb#L5

This could become:
`@person = person || Person.new`

Then your're always guaranteed to have a person model to operate on

spec/policies/presentation_policy_spec.rb
jacquesporveau: Interesting. OK I didn't know that I will make the changes.

app/policies/presentation_policy.rb
jacquesporveau: This how we are currently deciding if a person is signed in. We have different permissions for: not signed in, signed in and admin. If person is nil then they are not signed in.

qr8r: Sure, that does work. 

I think it would simplify things if you considered a person to not be signed in if they were unpersisted. That way you are always operating on the same object, and not sometimes nil and sometimes a person

qr8r (approved):

app/policies/presentation_policy.rb
jacquesporveau: For documentations sake the resolution to this conversation is listed here:

https://www.pivotaltracker.com/story/show/142711001

PR #56, opened 2017-03-27
Feature/inline errors

jacquesporveau:
https://www.pivotaltracker.com/story/show/140356551

We want to display errors in-line. I made a PR to do another way and SimpleForm was suggested. Sounds like we use that on a bunch of the projects we have going now so I decided to go with it.


app/views/presentations/_form.html.erb
```diff
@@ -1,36 +1,15 @@
...
-    <%= f.label :duration %><br>
-    <%= f.text_field :duration %> Minutes
-  </p>
-
-  <p>
-    <%= f.label :event %><br>
-    <%= f.collection_select(:event_id, Event.upcoming, :id, :name, prompt: true) %>
-  </p>
+  <%= f.input :topic %>
+  <%= f.input :description, as: :string %>
```
qr8r: do you need the as here? I would have expected simple form to have figured that on on its own

app/views/events/_form.html.erb
```diff
@@ -1,23 +1,8 @@
...
-    <%= f.label :date %><br>
-    <%= f.date_field :date %>
-  </p>
-
-  <p>
-    <%= f.submit %>
-  </p>
+  <%= f.input :name %>
+  <%= f.input :location %>
+  <%= f.input :date, as: :date %>
```
qr8r: do you have to specify here?

app/views/people/_form.html.erb
```diff
@@ -1,43 +1,12 @@
...
-    <%= f.submit %>
-  </p>
+<%= simple_form_for @person do |f| %>
+
+  <%= f.input :first_name %>
+  <%= f.input :last_name %>
+  <%= f.input :email %>
+  <%= f.input :password %>
+  <%= f.input :password_confirmation %>
+  <%= f.input :date_of_birth, as: :date %>
```
qr8r: Same

qr8r (commented):
@jacquesporveau Just had some questions about how you are using simpleform.

I kind of expected it to infer a lot of the input types you are specifying

app/views/presentations/_form.html.erb
jacquesporveau: This is actually a binary field so if I don't have as string is errors out.

jacquesporveau:
Requested changes made @brendandeere

app/views/presentations/_form.html.erb
natyork: That makes sense to me from the form perspective, but do you happen to know why we specified presentations description as a binary in our schema?

natyork (commented):
No changes from me - simple_form seems really interesting, good job getting it implemented. 

I just had the one question as to whether you happen to know why we defined presentation description as a binary in our schema?

jacquesporveau:
@natyork Descriptions might be long and I think the string field is limited to 255 char.

app/views/presentations/_form.html.erb
jacquesporveau: Descriptions might be long and I think the string field is limited to 255 char.

qr8r (approved):

PR #57, opened 2017-03-27
Feature/view presentations

natyork:
This PR creates a filter for the presentations index view, IOT filter page by presenters , per pivotal story: https://www.pivotaltracker.com/story/show/140356273

### DIFF
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -1,8 +1,10 @@
 class PresentationsController < ApplicationController
   def index
-    @presentations = Presentation.joins(:event)
-                                 .merge(Event.order(date: :desc))
-                                 .page(params[:page])
+    @search = Presentation.ransack(params[:search])
+    @presentations = @search.result(distinct: true)
+                            .joins(:event)
+                            .merge(Event.order(date: :desc))
+                            .page(params[:page])
   end
 
   def show
--- a/app/views/presentations/_filter.html.erb
+++ b/app/views/presentations/_filter.html.erb
@@ -1,0 +1,8 @@
+<%= search_form_for @search do |f| %>
+  <p>
+    <%= f.label 'Filter by Presenter:' %>
+    <%= f.collection_select(:person_id_eq, Person.all, :id, :first_name, prompt: true) %>
+    <%= f.submit %>
+    <%= link_to 'Clear Filter', presentations_path %>
+  </p>
+<% end %>
--- a/app/views/presentations/index.html.erb
+++ b/app/views/presentations/index.html.erb
@@ -1,5 +1,7 @@
 <h1>Presentations:</h1>
 
+<%= render 'filter' %>
+
 <table>
   <thead>
     <tr>
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,9 @@ gem 'kaminari'
 # Use Simpleform for forms
 gem 'simple_form'
 
+# Use Ransack for creating search forms
+gem 'ransack'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platform: :mri
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -158,6 +158,8 @@ GEM
       omniauth (~> 1.2)
     parser (2.4.0.0)
       ast (~> 2.2)
+    polyamorous (1.3.1)
+      activerecord (>= 3.0)
     powerpack (0.1.1)
     pry (0.10.4)
       coderay (~> 1.1.0)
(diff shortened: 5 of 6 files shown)


jacquesporveau:
jenkins test this please

app/controllers/presentations_controller.rb
```diff
@@ -1,8 +1,16 @@
 class PresentationsController < ApplicationController
   def index
-    @presentations = Presentation.joins(:event)
-                                 .merge(Event.order(date: :desc))
-                                 .page(params[:page])
+    if params[:presentation] && params[:presentation][:person_id].present?
```
jacquesporveau: Brendan has asked me to move to the safe navigation operator. 

http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

app/views/presentations/_filter.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= form_tag presentations_path, method: 'get' do %>
+  <p>
+  <%= label_tag('Filter by Presenter:') %>
```
jacquesporveau: Indent.

app/views/presentations/index.html.erb
```diff
@@ -23,6 +23,9 @@
   </tbody>
   <tfoot>
     <tr>
+      <td><%= render 'filter' %></td>
```
jacquesporveau: What about the idea of having this filter at the top somewhere. The table will differ in row count making the filter show up at different places down the page. If it is above all the rows it will always been in the same place.

jacquesporveau (changes requested):

app/views/presentations/_filter.html.erb
natyork: I think i need to look at my vim settings, because my indents often un-indent automatically if I make changes to the line for erb files....it must be interpreting the tags as being at the same level? I'll investigate, because you've caught this a couple times.

app/views/presentations/index.html.erb
natyork: good point, i like that idea

app/views/presentations/_filter.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= form_tag presentations_path, method: 'get' do %>
+  <p>
+    <%= label_tag('Filter by Presenter:') %>
+    <%= collection_select(:presentation, :person_id, Person.all, :id, :first_name, prompt: true) %>
```
graygilmore: Does this allow us to also set a "All Presenters" option?

graygilmore (changes requested):
The commit that adds the form cannot function without the controller changes. The commits should either be re-ordered or squashed into a single commit.

app/views/presentations/_filter.html.erb
natyork: I've added in the following option to collection_select:
 `include_blank: 'All Presenters' `
This allows us to filter for All presenters in the controller.

natyork:
Jenkins failing due to Rubocop Ruby parser defaulting to v 2.1. PR #55 updates the .rubocop.yml file to set TargetRubyVersion to 2.3.1. When #55 is merged, there should be no rubocop offenses for the safe navigation operator.

qr8r:
@natyork Since PR #55 is blocking future PRS, can someone extract the changes required to rubycop into their own pr that I can merge separately from the new features being added there?

app/controllers/presentations_controller.rb
qr8r: I would also take a look at the methods ruby hash probides you for doing this.

#fetch and #dig

qr8r:
This is great! It definitely works. I mentioned the other day there is a fairly common rails tool for implementing this. I was wondering if you had been able to find it.

natyork:
@brendandeere I did look at Ransack as an option, but I thought it might be more than we needed for this app. However, if you think that is a more sensible/flexible approach, I could use that instead!

qr8r:
Yeah, I think I would like to see Ransack implemented. Or at the very least if its not something we wanted to do now,  I would like to see the ActiveRecord methods moved out of the controller and into a scope on Presentation.

natyork:
@brendandeere, this PR has been updated to make use of Ransack for the filter feature.

app/controllers/presentations_controller.rb
```diff
@@ -1,8 +1,10 @@
 class PresentationsController < ApplicationController
   def index
-    @presentations = Presentation.joins(:event)
-                                 .merge(Event.order(date: :desc))
-                                 .page(params[:page])
+    @q = Presentation.ransack(params[:q])
```
jacquesporveau: I'd like this instance variable named a little bit more intuitively. I am guessing it means query but I don't know if that's what it is.

jacquesporveau (dismissed):
Would also like to see the commits broken up a little. Maybe there is an instance in the future where we want to keep the ransack gem installed but ditch the implementation. Things like that.

app/controllers/presentations_controller.rb
natyork: It's a ransack default search parameter. You can customize it, but I left the default for the time being to minimize query string length.
https://github.com/activerecord-hackery/ransack/tree/v1.8.2#simple-mode

graygilmore (approved):
Looks good on my end. Would love to know what @brendandeere thinks as my ransacking is not up to date.

natyork:
Split into two commits and updated the search key to something more readable per @jacquesporveau suggestions . Waiting on @brendandeere review before merging.

qr8r (approved):

PR #58, opened 2017-03-28
Feature/flash messages

natyork:
Create a view to render all flash messages per https://www.pivotaltracker.com/story/show/142562277

Flash messages will be displayed in a div with `class="alert alert-#{key}"` where key is the flash type, so flash messages can be styled by type.

There are no custom flash types currently defined, but if in future custom flash methods are defined, such as for https://www.pivotaltracker.com/story/show/142092195 they can be defined in the applications controller per: http://api.rubyonrails.org/classes/ActionController/Flash/ClassMethods.html

### DIFF
--- a/app/views/layouts/_flash-messages.html.erb
+++ b/app/views/layouts/_flash-messages.html.erb
@@ -1,0 +1,7 @@
+<% if flash.any? %>
+  <div class='alert-container'>
+    <% flash.each do |key, value| %>
+      <%= content_tag :div, value, class: "alert alert-#{key}" %>
+    <% end %>
+  </div>
+<% end %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/_flash-messages.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class='alert-container'>
```
graygilmore: Should we have a conditional here like `if flash.any?`?

graygilmore (changes requested):
We can squash the two commits into one as they don't make sense without each other.

jacquesporveau (approved):
LGTM besides Grays comment.

natyork:
Commits squashed and conditional rendering of flash messages added per @graygilmore comments.

qr8r (approved):

natyork:
@graygilmore has moved this story to the icebox. I am going to close this PR.

graygilmore (approved):

graygilmore:
Merging ❤️

PR #59, opened 2017-03-28
Feature/auto populate

jacquesporveau:
https://www.pivotaltracker.com/story/show/141823283

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -16,6 +16,10 @@ class Event < ApplicationRecord
     where('date = ?', Date.current - NOTIFICATION_PERIOD)
   end
 
+  def upcoming?
+    Event.upcoming.include?(self)
+  end
+
   def duration
     presentations.map(&:duration).sum
   end
--- a/app/views/presentations/_sign_up_link.html.erb
+++ b/app/views/presentations/_sign_up_link.html.erb
@@ -1,0 +1,5 @@
+<%=
+  if event&.upcoming?
+    link_to 'Sign up', new_presentation_path(prepopulated_event: event.id)
+  end
+%>
--- a/spec/models/event_spec.rb
+++ b/spec/models/event_spec.rb
@@ -121,4 +121,23 @@ RSpec.describe Event, type: :model do
       end
     end
   end
+
+  describe '#upcoming?' do
+    subject { event.upcoming? }
+    context 'the event is in the future' do
+      let!(:event) do
+        FactoryGirl.create(:event, date: 1.hour.from_now)
+      end
+
+      it { is_expected.to eq(true) }
+    end
+
+    context 'the event is in the past' do
+      let!(:event) do
+        FactoryGirl.create(:event, date: -1.hours.from_now)
+      end
+
+      it { is_expected.to eq(false) }
+    end
+  end
 end
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -12,6 +12,7 @@ class PresentationsController < ApplicationController
   def new
     @presentation = Presentation.new
     authorize @presentation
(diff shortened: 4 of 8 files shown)


spec/models/event_spec.rb
```diff
@@ -121,4 +121,23 @@
       end
     end
   end
+
+  describe '#upcoming?' do
+    subject(:event_upcoming?) { event.upcoming? }
```
graygilmore: If you're not going to use the named subject you can just do `subject { event.upcoming? }`.

spec/models/event_spec.rb
```diff
@@ -121,4 +121,23 @@
...
+    context 'the event is in the future' do
+      let!(:event) do
+        FactoryGirl.create(:event, date: Time.zone.today + 1.hour)
```
graygilmore: Instead of doing math you could do something like: `1.year.from_now`.

graygilmore: Can we get away with using `build` here? Might not get picked up by the scope, I guess.

app/views/presentations/_sign_up_link.html.erb
```diff
@@ -0,0 +1,5 @@
+<%=
+  if event&.upcoming?
+    link_to 'Sign up', new_presentation_path(prepopulated_event: event.id)
```
graygilmore, line 3: Instead of `prepopulated_event` could we just use `event`? Or even `event_id` as that's what we're passing.

graygilmore (changes requested):

spec/models/event_spec.rb
jacquesporveau: Can't use build as it won't be picked up by scope.

app/views/presentations/_sign_up_link.html.erb
jacquesporveau: Mmmmmm I kind of want there to be some reference to what this is for when it hits the controller.
This allows me to have `params[:prepopulated_event]` in my controller. I know what that is.
Having a reference to `params[:event_id]` in my presentations controller is kind of odd. Why do I have a reference to this? What is this being used for? It's a little bit unclear why it's there as the event is needed for a specific reason.

app/models/event.rb
```diff
@@ -16,6 +16,10 @@ class Event < ApplicationRecord
     where('date = ?', Date.current - NOTIFICATION_PERIOD)
   end
 
+  def upcoming?
+    Event.upcoming.include?(self)
```
qr8r, line 20: this could also be `self.class`

app/views/presentations/_sign_up_link.html.erb
qr8r: So I have a question that is meant more for discussion: 

Does it ever make sense to create a presentation that is not associated to a specific event? 
Is there ever a time when Im trying to create a presentation that I dont already know specifically which event I want it to be associated with?

jacquesporveau: In the current state of the application? No. In the future? Maybe.

qr8r: Cool, If it never makes sense to build a presentation in isolation form an event. Should presentations then be a nested resource of events?

jacquesporveau: I see where you are going with this and I like it.

app/models/event.rb
jacquesporveau: I think you are trying to lead me to: If we changed the name of the class in the future we wouldn't need to update this method. I am going to take the advice.

jacquesporveau:
I am going to be tackling this a completely different way so I am going to close this PR.

PR #61, opened 2017-03-29
Feature/auto populate

jacquesporveau:
https://www.pivotaltracker.com/story/show/141823283

### DIFF
--- a/app/controllers/presentations_controller.rb
+++ b/app/controllers/presentations_controller.rb
@@ -12,7 +12,8 @@ class PresentationsController < ApplicationController
   end
 
   def new
-    @presentation = Presentation.new
+    @event = Event.find_by(id: params[:event_id]) || Event.new
+    @presentation = @event.presentations.build
     authorize @presentation
   end
 
--- a/app/views/presentations/_form.html.erb
+++ b/app/views/presentations/_form.html.erb
@@ -4,7 +4,12 @@
   <%= f.input :topic %>
   <%= f.input :description, as: :string %>
   <%= f.input :duration, as: :numeric %>
-  <%= f.association :event, collection: upcoming_events, label_method: :display_name %>
+  <%= f.association :event,
+    collection: upcoming_events,
+    prompt: true,
+    selected: @event.id,
+    label_method: :display_name
+  %>
   <% if current_person && current_person.admin? %>
     <%= f.label :presenter %>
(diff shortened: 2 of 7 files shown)


app/controllers/presentations_controller.rb
```diff
@@ -11,6 +11,7 @@ def show
 
   def new
     @presentation = Presentation.new
+    @event = Event.find(params[:event_id]) if params[:event_id]
```
graygilmore: I think you can remove the conditional here. If `params[:id]` is `nil` then it won't be able to find an Event and `@event` ends up being `nil`, too.

* pretty sure

graygilmore (changes requested):

app/controllers/presentations_controller.rb
jacquesporveau: Actually it 500's. If we did a `.find_by(id: params[:event_id])` that would work but I don't want to set an instance variable to `nil` if I don't have to.

graygilmore (changes requested):
Looks like this is missing from the home page sign up link.

jacquesporveau:
Mmmmmm yep alright that was sloppy. I've gone through and done my best to nail all the spots that this new link should be in. Worth noting that we could probably use more sign up links throughout the app. I added a bit about that to the QA audit issue.

app/controllers/presentations_controller.rb
qr8r: In the next commit though you call `@event&.id` anyway though. when you call an instance variable it is created with the initial value of nil. 

i'd either to the `find_by` or something like `@event = Event.find_by(id: params[:event_id]) || Event.new` 
that way you're always operating on an event object. 

Id also consider always building the presentation off of the event object `@event.presentations.build`

jacquesporveau: Oh that`.build` is cool. I didn't know you can do that / or it worked that way.

qr8r: http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html

jacquesporveau: This comment is showing up as not updated but if you view the diff it actually has been changed.

qr8r (approved):

graygilmore (approved):

PR #62, opened 2017-03-30
Feature/request model

jacquesporveau:
https://www.pivotaltracker.com/story/show/140356337

I've decided to break this issue up into multiple PRs. This first PR is just about establishing the requested topic model.

### DIFF
--- a/spec/models/topic_request_spec.rb
+++ b/spec/models/topic_request_spec.rb
@@ -1,0 +1,5 @@
+require "rails_helper"
+
+RSpec.describe TopicRequest, type: :model do
+  it { is_expected.to validate_presence_of(:name) }
+end
--- a/app/models/topic_request.rb
+++ b/app/models/topic_request.rb
@@ -1,0 +1,3 @@
+class TopicRequest < ApplicationRecord
+  validates :name, presence: true
+end
--- a/db/migrate/20170329230309_create_topic_requests.rb
+++ b/db/migrate/20170329230309_create_topic_requests.rb
@@ -1,0 +1,9 @@
+class CreateTopicRequests < ActiveRecord::Migration[5.0]
+  def change
+    create_table :topic_requests do |t|
+      t.string :name, null: false
+
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170306230913) do
+ActiveRecord::Schema.define(version: 20170329230309) do
 
   create_table "events", force: :cascade do |t|
     t.date     "date",       null: false
@@ -76,6 +76,12 @@ ActiveRecord::Schema.define(version: 20170306230913) do
     t.datetime "created_at",         null: false
     t.datetime "updated_at",         null: false
     t.index ["person_id"], name: "index_staff_on_person_id"
+  end
+
+  create_table "topic_requests", force: :cascade do |t|
+    t.string   "name",       null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
   end
 
 end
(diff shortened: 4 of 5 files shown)


db/migrate/20170329230309_create_requested_topics.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRequestedTopics < ActiveRecord::Migration[5.0]
+  def change
+    create_table :requested_topics do |t|
```
graygilmore: Naming is Hard :tm:

What do you think about `TopicRequests`? Something about the past-tense `requested` bugs me. What do you think?

db/migrate/20170329230309_create_requested_topics.rb
```diff
@@ -0,0 +1,9 @@
...
+      t.string :name, null: false
```
graygilmore: Do you think `name` is the best thing we could go with?

```ruby
@requested_topic.name
```

```ruby
@topic_request.title
```

Maybe? `.subject`?

Naming is Hard :tm:

app/models/requested_topic.rb
```diff
@@ -0,0 +1,4 @@
+class RequestedTopic < ApplicationRecord
+  validates :name,
+            presence: true
```
graygilmore: No need to drop this to another line.

spec/models/requested_topic_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.describe RequestedTopic, type: :model do
+  describe 'name' do
+    subject { described_class.new(requested_topic) }
+    context 'when valid' do
```
graygilmore: I believe we added shoulda matchers to the project. This whole spec file can be:

```ruby
require "rails_helper"

RSpec.describe RequestedTopic, type: :model do
  it { is_expected.to validate_presence_of(:name) }
end
```

graygilmore (changes requested):

db/migrate/20170329230309_create_requested_topics.rb
jacquesporveau: I like TopicRequest. Now that I think about it having a past tense is kind of implying too much about the class and doesn't give us flexibility to do other things with it in the future.

db/migrate/20170329230309_create_requested_topics.rb
jacquesporveau: I wanted to go with `.topic` but I threw it around the pod and `.name` came out. It's tricky because we are naming a presentations topic prop as 'topic' so topic is kind of what I expect this thing to be named as. If we have a TopicRequest I think name and my suggestion of `topic_request.topic` are the only ones that really make sense to me. I think `topic_request.topic` kind of sucks because it's a prop name named after the class.

app/models/requested_topic.rb
jacquesporveau: Following previous style conventions from the rest of the models.

jacquesporveau:
```
# app/models/event.rb
  NOTIFICATION_PERIOD = 1.day
  MAXIMUM_DURATION = Figaro.env.maximum_event_duration.to_i
  has_many :presentations, dependent: :destroy

  validates :date,
            presence: true
  validates :location,
            presence: true
  validates :name,
            presence: true

  scope :upcoming, -> { where('date >= ?', Time.zone.today) }
  scope :requiring_notification, -> do
  end

  def duration
    presentations.map(&:duration).sum
  end
end
```

jacquesporveau: Gray and I have talked this over and come to the conclusion that it's OK to break past conventions if your pattern is better.

qr8r (approved):

spec/models/topic_request_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
+
+RSpec.describe TopicRequest, type: :model do
+  it { is_expected.to validate_presence_of(:name) }
```
graygilmore, line 4: Much nicer!

graygilmore (approved):

PR #63, opened 2017-03-30
Feature/request presentations

jacquesporveau:
Part two of https://www.pivotaltracker.com/story/show/140356337

### DIFF
--- a/app/controllers/topic_requests_controller.rb
+++ b/app/controllers/topic_requests_controller.rb
@@ -1,0 +1,5 @@
+class TopicRequestsController < ApplicationController
+  def index
+    @topic_requests = TopicRequest.all
+  end
+end
--- a/app/controllers/topic_requests_controller.rb
+++ b/app/controllers/topic_requests_controller.rb
@@ -1,5 +1,26 @@
 class TopicRequestsController < ApplicationController
   def index
     @topic_requests = TopicRequest.all
+    @topic_request = TopicRequest.new
+  end
+
+  def create
+    @topic_request = TopicRequest.new(topic_request_params)
+
+    if @topic_request.save
+      flash[:success] = 'Topic request was successfully created.'
+      redirect_to requests_presentations_path
+    else
+      @topic_requests = TopicRequest.all
+      render 'index'
+    end
+  end
+
+  private
+
+  def topic_request_params
+    params.require(:topic_request).permit(
+      :name
+    )
   end
 end
--- a/app/views/topic_requests/_form.html.erb
+++ b/app/views/topic_requests/_form.html.erb
@@ -1,0 +1,6 @@
+<%= simple_form_for topic_request do |f| %>
+
+  <%= f.input :name %>
+  <%= f.button :submit %>
+
+<% end %>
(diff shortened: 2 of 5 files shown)


app/controllers/topic_requests_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  def index
+    @topic_requests = TopicRequest.all
+    @topic_request = TopicRequest.new
+  end
+
+  def create
+    topic_request = TopicRequest.new(topic_request_params)
+    authorize topic_request
+
+    flash[:notice] = if topic_request.save
```
graygilmore: Should we do a `flash[:success]` and `flash[:error]` instead?

app/controllers/topic_requests_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+                       'Topic request was succesfully created.'
+                     else
+                       'Error in creating topic request.'
+                     end
+    redirect_to requests_presentations_path
```
graygilmore: If we do a redirect here do we get specific error messages for why it wasn't submitted? Like if the user leaves it blank.

app/policies/topic_request_policy.rb
```diff
@@ -0,0 +1,5 @@
+class TopicRequestPolicy < ApplicationPolicy
+  def create?
+    true
```
graygilmore: Should we skip this entirely and just not do `authorize topic_request`?

spec/policies/topic_request_policy_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe TopicRequestPolicy do
+  let(:topic_request) { FactoryGirl.build(:topic_request) }
+  context 'the person is not signed in' do
+    subject { described_class.new(nil_person, topic_request) }
+    let(:nil_person) { nil }
+
+    describe '#create?' do
```
graygilmore: I think this spec file is a little backwards. Normally you'd see something like:

```ruby
describe "#my_method" do
  subject { my_method }

  it "does something" do
  end

  context "context one" do
    it "does another thing" do
    end
  end

  context "context two" do
    it "does something else" do
    end
  end
end
```

graygilmore (changes requested):

app/controllers/topic_requests_controller.rb
jacquesporveau: Yes it did do that.

app/policies/topic_request_policy.rb
jacquesporveau: Yes we should. Done now.

spec/policies/topic_request_policy_spec.rb
jacquesporveau: Point taken. This code has all been removed along with the policy it was testing though.

natyork (commented):
Looks good to me :) Nice work!

app/controllers/topic_requests_controller.rb
jacquesporveau: :+1:

graygilmore (approved):

PR #64, opened 2017-03-30
Bug/presentation sign up

jacquesporveau:
https://www.pivotaltracker.com/story/show/141187241/comments/167429831

### DIFF
--- a/spec/presenters/presentation_presenter_spec.rb
+++ b/spec/presenters/presentation_presenter_spec.rb
@@ -1,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe PresentationPresenter do
+  subject { described_class.new(presentation, nil) }
+
+  describe '#associable_events' do
+    let(:event) { FactoryGirl.build(:event) }
+    let(:past_event) { FactoryGirl.build(:event, date: 1.day.ago) }
+    context 'the presentation has an associated event' do
+      context 'the event is in the future' do
+        let(:presentation) do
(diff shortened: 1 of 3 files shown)


spec/presenters/presentation_presenter_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe PresentationPresenter do
+  subject { described_class.new(presentation, nil) }
+
+  describe '#associable_events' do
```
qr8r, line 6: This should probably also test that past events arent included

qr8r (changes requested):
:+1:  other than minor test comment. Also #NamingIsHard

spec/presenters/presentation_presenter_spec.rb
jacquesporveau: Sounds good. I was thinking about testing that but my thought pattern was that the `.upcoming` method was already tested so I could skip that test. I'll make the changes.

qr8r: That makes an assumption about the internal implementation of this method. Really I dont care if you use the upcoming scope or some other mechanism for filtering Events.

But what I care about testing is behaviour: 
- No past events are included 
- The currently associated event is present in the list

jacquesporveau: Alright I've updated the spec now.

qr8r (approved):

PR #65, opened 2017-03-30
Create custom 403 error page for Pundit errors

natyork:
Rescue pundit errors and redirect to /403 route.

https://www.pivotaltracker.com/story/show/142092195

### DIFF
--- a/app/views/errors/forbidden.html.erb
+++ b/app/views/errors/forbidden.html.erb
@@ -1,0 +1,7 @@
+<div class='main'>
+  <h1>Woops!</h1>
+  <p>You do not have access to this page.</p>
+  <% if !current_person %>
+    <p>  This is most likely being caused because you are not <%= link_to 'signed in', '/auth/google_oauth2' %>.</p>
+  <% end %>
+</div>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,6 +3,7 @@ Rails.application.routes.draw do
   get 'auth/failure', to: redirect('/')
   get 'signout', to: 'sessions#destroy', as: 'signout'
   get 'authentication/google', to: 'authentication#google'
+  match '/403', to: 'errors#forbidden', via: :all
 
   resources :sessions, only: [:destroy]
   resources :people
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,0 +1,5 @@
+class ErrorsController < ApplicationController
+  def forbidden
+    render status: 403
+  end
+end
--- a/config/application.rb
+++ b/config/application.rb
@@ -11,5 +11,11 @@ module StaffTracker
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+
+    config.action_dispatch
+          .rescue_responses["Pundit::NotAuthorizedError"] = :forbidden
+
+    require Rails.root.join('lib/custom_public_exceptions')
+    config.exceptions_app = CustomPublicExceptions.new(Rails.public_path)
   end
 end
--- a/lib/custom_public_exceptions.rb
+++ b/lib/custom_public_exceptions.rb
@@ -1,0 +1,10 @@
+class CustomPublicExceptions < ActionDispatch::PublicExceptions
+  def call(env)
+    status = env["PATH_INFO"][1..-1]
+    if status == '403'
+      Rails.application.routes.call(env)
+    else
+      super
+    end
+  end
+end
--- a/spec/controllers/errors_controller_spec.rb
+++ b/spec/controllers/errors_controller_spec.rb
@@ -1,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe ErrorsController, type: :controller do
+  describe 'GET #forbidden' do
+    subject { get :forbidden }
+    it { is_expected.to render_template :forbidden }
+    it { is_expected.to have_http_status(403) }
+  end
+end
--- a/spec/lib/custom_public_exceptions_spec.rb
+++ b/spec/lib/custom_public_exceptions_spec.rb
@@ -1,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe CustomPublicExceptions do
+  describe '#call' do
+    subject { described_class.new(Rails.public_path).call(env) }
+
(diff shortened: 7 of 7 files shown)


config/routes.rb
```diff
@@ -4,6 +4,8 @@
   get 'signout', to: 'sessions#destroy', as: 'signout'
   get 'authentication/google', to: 'authentication#google'
 
+  get '/403', to: 'errors#forbidden', as: 'forbidden'
```
jacquesporveau: Probably don't need the whitespace above.

spec/controllers/events_controller_spec.rb
```diff
@@ -105,7 +105,7 @@
       let(:params) { FactoryGirl.attributes_for(:event) }
 
       it 'throws a not authorized error' do
-        expect { subject }.to raise_error(Pundit::NotAuthorizedError)
+        expect(subject).to redirect_to forbidden_path
```
jacquesporveau: You could leave these specs in and write more for the redirect. We are still raising the error but now we are rescuing it.

app/views/errors/forbidden.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class='main'>
+  <h1>Woops!</h1>
+  <p>You do not have access to this page.</p>
+  <p>  Are you <%= link_to 'signed in', '/auth/google_oauth2' %>?</p>
```
jacquesporveau: I feel like a tool for bringing this up but I think the question mark should be with the signed in. :nervous_smiley:

jacquesporveau (commented):
Just suggestions from me.

spec/controllers/events_controller_spec.rb
natyork: As discussed, rescuing the error results in the raise_error check to fail. @brendandeere, do you have any suggestions for this test?

app/controllers/application_controller.rb
```diff
@@ -3,9 +3,17 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   helper_method :current_person
 
+  rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
```
qr8r: https://makandracards.com/makandra/41790-how-rails-chooses-error-pages-404-500-for-exceptions
https://makandracards.com/makandra/12807-custom-error-pages-in-rails-3-2+
http://guides.rubyonrails.org/layouts_and_rendering.html#the-status-option

qr8r (changes requested):

app/views/errors/forbidden.html.erb
natyork: I'd be interested to know what @graygilmore thinks. I thought it looked a little funny with the punctuation in the link, but that's just my opinion - I'm not set on one way in particular. 
**Screenshots:**
https://drive.google.com/open?id=0B69-mH0H4uz4S0wtV2Z0T1pLd2M  vs https://drive.google.com/open?id=0B69-mH0H4uz4RE9IS0VvSUR0VkE

spec/controllers/events_controller_spec.rb
qr8r: If the error is no longer being thrown in the tests, then there is no reason to test for it

graygilmore: I agree with Brendan's sentiment above. Furthermore... if we're creating a common case for all auth errors (redirecting to a 403) we could create a [shared example](https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/example-groups/shared-examples) and replace all of our `raise_error` tests with something like:

```ruby
context "when a user does not have permissions" do
  # setup bad permissions here
  include_examples "failed authorization"
end
```

I'd look at doing something like this if we're checking for `Pundit::NotAuthorizedError` in several tests.

app/views/errors/forbidden.html.erb
graygilmore: Should we just conditionally show the link?

```
<% if !current_user %>
  <p>This is most likely being caused because you are not <a href="/auth/google_oauth2">signed in</a>.
<% end %>
```

If you want to keep it as it is now the question mark would probably remain outside of the link.

app/controllers/application_controller.rb
natyork: Originally I had set up the custom error page per:
https://thepugautomatic.com/2014/08/404-with-rails-4/ 
and updated the application config file to include ```config.action_dispatch.rescue_responses["Pundit::NotAuthorizedError"] = :forbidden``` per:
https://github.com/elabs/pundit#rescuing-a-denied-authorization-in-rails

This allowed us to keep the default static pages as well as use a custom dynamic error page.

However when the error was raised there was no redirect to the 403 route, and the template was rendered on the same page where the error was raised. 

Using this method, gave a redirect, and removed some of the complexity  in setting up the custom dynamic error page.  It does however add complexity to the application controller and if we also wanted to add other custom error pages, it may need to be changed back to the above approach anyway.

I will add another commit to use the original approach so you can compare them side-by-side and we can discuss the pros/cons of each approach.

spec/controllers/events_controller_spec.rb
natyork: @brendandeere & @jacquesporveau, after updating the approach used to display the error page, the `Pundit::NotAuthorizedError` is raised and can be tested. 

Agree with @graygilmore that the shared example approach would be a great way to DRY up our specs. Since this pattern was established earlier in the project, I think this could be a separate chore on our pivotal board, rather than something that gets changed for this PR. I have added to pivotal icebox!

natyork:
Per discussion with @brendandeere , I have added a second approach for rendering the custom error page (OPTION 2). In order to locally view the custom error messages rather than the stack traces we see by default, I have created a temporary commit that updates the development config, in case you wish to do any local testing of this feature. Once we have determined the desired approach for the error page, I will clean up commits and rebase as required.

natyork:
Per discussion with @brendandeere, I have implemented the error page approach per https://thepugautomatic.com/2014/08/404-with-rails-4/ that allows creation of custom dynamic error pages while still maintaining the ability to use static pages, where desired.   

Note: if you wish to test locally, update ```config/environments/development.rb``` so that ```config.consider_all_requests_local = false```. I did not include this in the commit as it will override the helpful development stack traces you get for any error.

qr8r (approved):

PR #67, opened 2017-03-31
[WIP] - Feature/admin interface

jacquesporveau:
UPDATE: Brendan and I are going to try and introduce nested_attributes later on. Until then this doesn't need anymore review.

https://www.pivotaltracker.com/story/show/142421627

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,0 +1,10 @@
+class AdminController < ApplicationController
+  before_action :authenticate_admin
+
+  private
+
+  def authenticate_admin
+    raise Pundit::NotAuthorizedError unless current_person
+    authorize current_person, :admin?
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,5 +12,11 @@ Rails.application.routes.draw do
   end
   resources :presentations, except: [:edit]
 
+  namespace :admin do
+    resources :people, only: [:index] do
+      put 'batch_update', to: 'people#batch_update', on: :collection
+    end
+  end
+
   root 'welcome#index'
 end
--- a/spec/controllers/admin_controller_spec.rb
+++ b/spec/controllers/admin_controller_spec.rb
@@ -1,0 +1,36 @@
+require 'rails_helper'
+
+RSpec.describe AdminController, type: :controller do
+  controller do
+    def authenticate_admin_mock
+      authenticate_admin
+    end
+  end
+
+  describe '#authenticate_admin' do
+    subject { controller.authenticate_admin_mock }
+
+    context 'the current person is not signed in' do
+      before { mock_pundit_user_as(nil) }
+      it 'raises a pundit not authorized error' do
+        expect { subject }.to raise_error(Pundit::NotAuthorizedError)
+      end
+    end
+
+    context 'the current person is a non admin' do
+      before { mock_pundit_user_as(non_admin) }
+      let(:non_admin) { FactoryGirl.build(:person) }
+      it 'raises a pundit not authorized error' do
+        expect { subject }.to raise_error(Pundit::NotAuthorizedError)
+      end
+    end
+
+    context 'the current person is an admin' do
+      before { mock_pundit_user_as(admin) }
+      let(:admin) { FactoryGirl.build(:person, :admin) }
+      it 'returns true' do
+        expect(subject).to eq(true)
+      end
+    end
+  end
+end


natyork (commented):
Agree with your comment that wherever you are raising `Pundit::NotAuthorizedError` the custom 403 message will be rendered.

app/policies/person_policy.rb
```diff
@@ -0,0 +1,5 @@
+class PersonPolicy < ApplicationPolicy
+  def admin?
+    person.present? && person.admin?
```
graygilmore: What's `person` here? Is this a global method that we have access to? Should it be `current_person`?

spec/controllers/admin_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+
+RSpec.describe AdminController, type: :controller do
+  describe '#authenticate_admin' do
```
graygilmore: Testing a private method is a bit of a code smell. The commit for this could use some more words, too!

spec/policies/person_policy_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe PersonPolicy do
+  describe '#admin?' do
+    context 'the person is not signed in' do
+      subject { described_class.new(nil_person, 'any object') }
```
graygilmore: Since we don't every test for `nil_person` we can just use `nil` and get rid of the `let` altogether.

spec/policies/person_policy_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      let(:nil_person) { nil }
+      it 'returns false' do
+        expect(subject.admin?).to eq(false)
+      end
+    end
+
+    context 'the person is a non admin' do
+      subject { described_class.new(non_admin, 'any_object') }
```
graygilmore: When we're describing a method it's usually nicer to set the subject once and then chance something particular about it in each context.

```ruby
describe "#admin?" do
  subject { described_class.new(person, "object").admin? }
  
  context "when the person is nil" do
    let(:person) { nil }
    it { is_expected.to eq false }
  end

  context "when the person is signed" do
    context "and is not an admin" do
      let(:person) {}
      it { is_expected.to eq false }
    end

    context "and is an admin" do
      let(:person) {}
      it { is_expected.to eq true }
    end
  end
end
```

Additionally, you'll note that my subject in the example above is the method call itself. That way you don't have to do `subject.admin?` in every assertion.

graygilmore (changes requested):
I'm not sure that this PR makes sense on its own, particularly with the namespace addition to the routes file.

app/policies/person_policy.rb
jacquesporveau: This is some built in Pundit stuff

```
class ApplicationPolicy
  attr_reader :person, :record

  def initialize(person, record)
    @person = person
    @record = record
  end
#...
```

If we trace person through all the loops and hoops it is essentially `current_person`

spec/policies/person_policy_spec.rb
jacquesporveau: Much cleaner thank you.

jacquesporveau:
Regarding Grays comment above. I am going to be adding admin namespaced controllers in future PR's. If I included all my work in one PR it would be like 300 lines so I tried to split it up.

spec/controllers/admin_controller_spec.rb
jacquesporveau: Yeah this controller doesn't have any actions so if I want 100% code coverage I think I need to test this.

graygilmore: I guess if the controller doesn't have any actions, why is it here?

jacquesporveau: It's for other controllers to inherit from. Application controller doesn't have any actions either.

jacquesporveau:
Gray and I have come to the conclusion that the first PR that was opened didn't really make sense to be split of from the second. The first was lacking context. I've merged in my other PR into this one. It makes for a long PR but there is more context of why I am taking the actions in my commits.

config/routes.rb
```diff
@@ -12,5 +12,11 @@
   end
   resources :presentations, except: [:edit]
 
+  namespace :admin do
```
graygilmore, line 15: Re 246674828a043517ebbf980c27f19e709d7a2a9f commit message. We like to follow the pattern of "What changed, then why it changed". Anybody looking at the code of the commit can see the changes that were made. Something a little more ideal might look like:

```
Add admin namespaced People routes

We want to create an admin area where people's roles can be updated. We're going
to namespace this behind `/admin` to make it clear that this is a separate entity
from the main site.
```

app/controllers/admin/people_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Admin
```
graygilmore: Another commit message comment.

ea7b8d5cbf0173da8af7208d9cec073a20906a47

> Also added a view for the sake of being able to write specs

Remember that each commit is its own thing. I don't necessarily know what came before this so "also" doesn't mean anything to me in this context.

> being able to write specs

This isn't a good enough reason to have a commit. These specs can be written when the actual view is created.

app/controllers/admin/people_controller.rb
```diff
@@ -3,5 +3,32 @@ class PeopleController < AdminController
     def index
       @people = Person.all
     end
+
+    def batch_update
+      changes_made = false
```
graygilmore: This has a bit of a smell to it. I'd recommend taking another look at how this is being implemented. If you're struggling with it fire me a message.

graygilmore (changes requested):

app/controllers/admin/people_controller.rb
jacquesporveau: Mmmmmm if I wrote those specs once I added the view then this commit would be under 100% coverage. I am aiming to have every single one of my commits at 100% coverage.

The alternative is to have the view commit fixup'd into this one but I feel like that's too much work stuffed into one commit.

graygilmore: I think the markup should be included in this commit. Without the markup, what is this commit doing?

jacquesporveau: The commit is adding in a controller action. A blank view is required to test that action without it throwing a template error. I was to isolate the commit to just the controller action.

app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,10 @@
+class AdminController < ApplicationController
+  before_action :authenticate_admin
+
+  private
+
+  def authenticate_admin
+    raise Pundit::NotAuthorizedError unless current_person
```
qr8r, line 7: Should we have to manually raise a pundit error, or is this something that pundit should be doing for us? 

...er I guess this is a guard so that we aren't passing `nil` to the policy?

spec/controllers/admin_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+RSpec.describe AdminController, type: :controller do
+  controller do
+    def authenticate_admin_mock
+      authenticate_admin
+    end
+  end
+
+  describe '#authenticate_admin' do
+    subject { controller.authenticate_admin_mock }
```
qr8r, line 11: Since this is a controller spec, would it make more sense to treat the subject as a request? 

`get authenticate_admin` or something.

app/controllers/admin/people_controller.rb
qr8r: I am :+1: for testing the controller action independently from adding markup to the view.

app/views/admin/people/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      <th>Name</th>
+      <th>Admin</th>
+    </tr>
+    <% @people.each do |person| %>
+      <tr>
+        <td><%= person.first_name %></td>
+        <td>
+          <%= simple_fields_for person do |f| %>
+            <%= f.input :id, as: :hidden, input_html:
+              { name: "people[person-#{person.id}]id" } %>
```
qr8r: :( Do we really need to hardcode the form names in?

app/models/person.rb
```diff
@@ -28,4 +28,14 @@ class Person < ApplicationRecord
   def admin?
     roles.detect(&:admin?).present?
   end
+
+  def toggle_admin!
```
qr8r: _OPINION_ 
You don't need to change this since this is just my opinion, but I want to raise the point to have the discussion.

Why is this a bang method? Does the public api provide a 'safe' alternative that you can call instead? Does this method do what it semantically says its doing.

I super don't like using bang methods unless there is an equivalent 'safe' method that the developer could be choosing instead. Otherwise I don't understand the semantics of the `!`. Almost all bang method in the Rubylang spec have an equivalent "safe" method. So the bang is semantically telling you that there is a "safer" choice.

Some people use the argument "This method is changing data", but to me that just means you need to give your method a descriptive name. `Hash#delete` doesn't take a bang and it modifies data directly. But the effect of the method is exactly what you expect from a method called delete. 
When you give your method a name that describes what it does, you dont need to add a bang to give your opinion that you think its a dangerous method.
`def my_super_scary_delete_everything_method_with_a_bang_so_everyone_knows_its_scary!` 

TL;DR I dont like using the bang to indicate your opinion that code is dangerous, that the method changes data, or that it raises an error. I only support its use to indicate that there is a safer alternative method you could be choosing

app/views/admin/people/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+            <%= f.input :admin?, as: :boolean, label: false, input_html:
```
qr8r: Not sure `admin?` is a valid field name..... what does this look like when its gets sent as params

app/views/admin/people/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <table>
+    <tr>
+      <th>Name</th>
+      <th>Admin</th>
+    </tr>
+    <% @people.each do |person| %>
+      <tr>
+        <td><%= person.first_name %></td>
+        <td>
+          <%= simple_fields_for person do |f| %>
```
qr8r: why did we decide not to do this as nested attributes for PersonRoles again?

app/controllers/admin/people_controller.rb
```diff
@@ -3,5 +3,32 @@ class PeopleController < AdminController
...
+      people_params.each do |param|
+        person = Person.find(param[:id])
```
qr8r: We have all the person ids at the start, any way we can preload all these records? 
Remember that in production each db request can take up to 200ms.

Another performance improvement we could bring here would be to wrap this whole loop with an ActiveRecord transaction (ill let you read up on that)

app/controllers/admin/people_controller.rb
```diff
@@ -3,5 +3,32 @@ class PeopleController < AdminController
...
+      redirect_to admin_people_path
+    end
+
+    private
+
+    def people_params
+      params.require(:people).map do |_person, param|
+        {
+          id: param[:id],
+          admin: param[:admin] == '0' ? false : true
```
qr8r: AR should be able to parse this out on its own

app/controllers/admin/people_controller.rb
```diff
@@ -3,5 +3,32 @@ class PeopleController < AdminController
...
+      else
+        flash[:error] = 'No changes to admin users were made'
+      end
+      redirect_to admin_people_path
+    end
+
+    private
+
+    def people_params
+      params.require(:people).map do |_person, param|
```
qr8r: Strong params should allow you to do this with permit

app/controllers/admin/people_controller.rb
```diff
@@ -3,5 +3,32 @@ class PeopleController < AdminController
...
+
+    def batch_update
+      changes_made = false
+      people_params.each do |param|
+        person = Person.find(param[:id])
+        next if person.admin? == param[:admin]
+        person.toggle_admin!
+        changes_made = true
+      end
+      if changes_made
```
qr8r: I don't know that we care that changes are made or not. I would only care if the `save` failed

qr8r (changes requested):

app/controllers/admin_controller.rb
jacquesporveau: Yeah can't call a policy on `nil`

jacquesporveau:
Going to open up another PR for a different branch / different approach.

PR #71, opened 2017-04-04
Update authentication controller redirect

natyork:
Pivotal Story: https://www.pivotaltracker.com/story/show/142616283

Adding ```redirect_to request.env['omniauth.origin']``` will redirect user to back to the page they were on prior to signing in.

https://github.com/omniauth/omniauth/wiki/saving-user-location

### DIFF
--- a/spec/controllers/authentication_controller_spec.rb
+++ b/spec/controllers/authentication_controller_spec.rb
@@ -10,14 +10,33 @@ RSpec.describe AuthenticationController, type: :controller do
       post :google
       response
     end
+
     it 'creates a new person' do
       expect { subject }.to change { Person.count }.by(1)
     end
-    it 'redirects the person to the root url' do
-      expect(subject).to redirect_to(root_url)
-    end
+
     it 'creates a new session' do
       expect { subject }.to change { session[:person_id] }.from nil
+    end
+
+    context 'omniauth origin is nil' do
+      before do
+        request.env['omniauth.origin'] = nil
+      end
+
+      it 'redirects the person to the root url' do
+        expect(subject).to redirect_to(root_url)
+      end
+    end
+
+    context 'omniauth origin is not nil' do
+      before do
+        request.env['omniauth.origin'] = 'http://test.host/presentations'
+      end
+
+      it 'redirects the person to the path location saved prior to login' do
+        expect(subject).to redirect_to(presentations_path)
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/controllers/authentication_controller_spec.rb
```diff
@@ -13,9 +13,17 @@
     it 'creates a new person' do
       expect { subject }.to change { Person.count }.by(1)
     end
-    it 'redirects the person to the root url' do
+
+    it 'redirects the person to the root url if omniauth origin is nil' do
```
jacquesporveau: the `if omniauth origin...` sounds like a context to me. Setting things in the it block maybe not so good.

spec/controllers/authentication_controller_spec.rb
```diff
@@ -13,9 +13,17 @@
...
+      request.env['omniauth.origin'] = nil
       expect(subject).to redirect_to(root_url)
     end
+
+    it 'redirects the person to the referer prior to login' do
```
jacquesporveau: Your other context.

natyork:
Updated per @jacquesporveau comments.

spec/controllers/authentication_controller_spec.rb
```diff
@@ -13,11 +13,29 @@
     it 'creates a new person' do
       expect { subject }.to change { Person.count }.by(1)
     end
-    it 'redirects the person to the root url' do
-      expect(subject).to redirect_to(root_url)
-    end
+
```
jacquesporveau, line 17: Notice you are putting whitespace above your it blocks. If you are going to adopt a different formatting style than what the file was previously following can you please update all the other occurrences throughout the file?

hint - *lines 12 - 13*

natyork: In general throughout all of our other specs, we have been including whitespace above it blocks. For this particular file, I did not add whitespace above the existing it blocks because I try not to make too many 'cleanup' changes outside the scope of the PR, as I find it can make the intention behind changes unclear. In this case, since it is one line and we have some context to the change, I agree that it is a good idea to include.

natyork:
Updated per @jacquesporveau comments

spec/controllers/authentication_controller_spec.rb
graygilmore: White space between `it` blocks 👍

qr8r (approved):

PR #72, opened 2017-04-05
Feature/nested attributes

jacquesporveau:
https://www.pivotaltracker.com/story/show/142421627

Update: Moving on to another project today. I will keep an eye for requested changes and perhaps make them on a lunch break or something but this one may sit a while.

### DIFF
--- a/app/controllers/admin/people_controller.rb
+++ b/app/controllers/admin/people_controller.rb
@@ -1,0 +1,7 @@
+module Admin
+  class PeopleController < AdminController
+    def index
+      @people = Person.all
+    end
+  end
+end
--- a/app/controllers/admin/people_controller.rb
+++ b/app/controllers/admin/people_controller.rb
@@ -3,5 +3,18 @@ module Admin
     def index
       @people = Person.all
     end
+
+    def batch_update
+      if Person.update(person_params.keys, person_params.values)
+        flash[:success] = 'People successfully updated.'
+      end
+      redirect_to admin_people_path
+    end
+
+    private
+
+    def person_params
+      params.require(:people).permit(person: [role_ids: []])[:person]
+    end
   end
 end
--- a/spec/controllers/admin_controller_spec.rb
+++ b/spec/controllers/admin_controller_spec.rb
@@ -1,0 +1,36 @@
+require 'rails_helper'
+
+RSpec.describe AdminController, type: :controller do
+  controller do
+    def authenticate_admin_mock
+      authenticate_admin
+    end
+  end
+
+  describe '#authenticate_admin' do
+    subject { controller.authenticate_admin_mock }
+
+    context 'the current person is not signed in' do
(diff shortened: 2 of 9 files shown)


app/controllers/admin/people_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+      if Person.update(person_params.keys, person_params.values)
+        flash[:success] = 'People successfully updated.'
+      end
+      redirect_to admin_people_path
+    end
+
+    private
+
+    def person_params
+      params.require(:people).permit(person: [role_ids: []])[:person]
```
graygilmore, line 17: 💯 🔥

spec/controllers/admin_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+
+RSpec.describe AdminController, type: :controller do
+  describe '#authenticate_admin' do
```
graygilmore: I believe that the controllers that are actually executing this path should cover these tests already. @brendandeere what do you think about this testing a private method?

spec/policies/person_policy_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe PersonPolicy do
+  describe '#admin?' do
+    subject { described_class.new(person, 'any object').admin? }
+    context 'the person is not signed in' do
+      let(:person) { nil }
+      it 'returns false' do
```
graygilmore: Could rewrite as `it { is_expected.to eq false }` if you want to shorten these.

graygilmore (changes requested):
Holy moly this is such a great improvement. Awesome job.

spec/controllers/admin_controller_spec.rb
jacquesporveau: I've changed how this test was to implemented. The test now creates a mock controller that inherits from the admin controller. Then the tests test that controller.

graygilmore: I think I'd still like Brendan's opinion. Testing a private method is counter to Stembolt convention.

graygilmore: After talking to Brendan about this we'd like to make use of RSpec's shared examples. What we really care about here is that the controllers that inherit from `AdminController` contain that functionality.

We don't want to have to write the same specs out every time for each controller that inherits from `AdminController`, though. Fortunately shared examples let's us define a series of assertions for the admin controller and simply include them in each controller that inherits from it.

```ruby
RSpec.describe SomeController do
  it_behaves_like "an admin controller"
end

RSpec.describe SomeOtherController do
  it_behaves_like "an admin controller"
end
```

jacquesporveau:
jenkins test this please

jacquesporveau:
jenkins test this please

PR #74, opened 2017-04-05
Refactor Pundit Policies

jacquesporveau:
https://www.pivotaltracker.com/story/show/142711001

I've refactored the pundit policies to all always
be operating on a person object. Because of this change
I also wrote a signed_in? method to improve readability.

Knowing we are always operating on a Person object in our
policies let's us clean up our code a bit and not have to consistenly
check for nil / safe navigation.

## Developer Notes

 - If #72 is merged before this we will also need to update the person policy to use the new style introduced in this PR.

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -2,12 +2,12 @@ class ApplicationPolicy
   attr_reader :person, :record
 
   def initialize(person, record)
-    @person = person
+    @person = person || Person.new
     @record = record
   end
 
   def create?
-    person.present? && person.admin?
+    signed_in?(person) && person.admin?
   end
 
   def new?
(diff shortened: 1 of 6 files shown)


app/policies/application_policy.rb
```diff
@@ -2,27 +2,33 @@ class ApplicationPolicy
   attr_reader :person, :record
 
   def initialize(person, record)
-    @person = person
+    @person = person || Person.new
```
graygilmore, line 5: This seems weird to me. We'll never end up creating this record. @brendandeere what do you think about this?

jacquesporveau: https://github.com/StemboltHQ/staff_tracker/pull/55#pullrequestreview-29768380

graygilmore (approved):

PR #75, opened 2017-04-06
Change descriptions data type in presentations table.

jacquesporveau:
https://www.pivotaltracker.com/story/show/142639017

### DIFF
--- a/db/migrate/20170406154040_change_data_type_for_description_in_table_presentations.rb
+++ b/db/migrate/20170406154040_change_data_type_for_description_in_table_presentations.rb
@@ -1,0 +1,5 @@
+class ChangeDataTypeForDescriptionInTablePresentations < ActiveRecord::Migration[5.0]
+  def change
+    change_column :presentations, :description, :text
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20170406154040_change_data_type_for_description_in_table_presentations.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDataTypeForDescriptionInTablePresentations < ActiveRecord::Migration[5.0]
+  def change
+    change_column :presentations, :description, :text
```
graygilmore, line 3: Do we lose any data with this migration?

jacquesporveau: http://sqlite.org/datatype3.html

If you search for type affinity on that website it mentions that it doesn't actually distinguish between blob and text column types. So on a SQLite level we haven't changed anything. 

If we move to PSQL which if this ever got launched I imagine we would then this would be the correct column data type.

To answer your question, no.

graygilmore (approved):

